PR #125, opened 2015-01-04
BMP180/BMP085 Barometric Pressure and Temperature Sensor

hstroh:
I've implemented a new protocol for sensor BMP180 (and its predecessor BMP085) which provides barometric pressure in hPa and temperature in °C. The algorithm for pressure and temperature measurement is based on the specification in the [official datasheet](https://ae-bst.resource.bosch.com/media/products/dokumente/bmp180/BST-BMP180-DS000-09.pdf). The implementation has been successfully tested using the sample values provided in the referenced sheet. 

I've added an option to display (barometric) pressure in the web UI for weather protocols. Pressure is displayed iff option `show-pressure` (which is a GUI setting) is set to `1`.The value to display is provided via option `pressure`. For proper display I've provided a new icon `pressure.png` which is a modification of `humi.png`.


CurlyMoo:
You need to remove the protocol_header and protocol_init.h files.

hstroh:
Done, files `protocol_header.h` and `protocol_init.h` have been removed.

libs/protocols/bmp180.c
```diff
@@ -0,0 +1,402 @@
...
+
+	protocol_register(&bmp180);
+	protocol_set_id(bmp180, "bmp180");
+	protocol_device_add(bmp180, "bmp180", "I2C Barometric Pressure and Temperature Sensor");
+	protocol_device_add(bmp180, "bmp085", "I2C Barometric Pressure and Temperature Sensor");
+	bmp180->devtype = WEATHER;
+	bmp180->hwtype = SENSOR;
+
+	options_add(&bmp180->options, 'i', "id", OPTION_HAS_VALUE, DEVICES_ID, JSON_STRING, NULL, "0x[0-9a-f]{2}");
+	options_add(&bmp180->options, 'o', "oversampling", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *) 1,
```
CurlyMoo: What does this do?

CurlyMoo:
1. You forgot the webgui update routine.
2. Remove the help function because this protocol doesn't support sending.
3. Your linked list implementation is not correct. Please notice that pilight starts a new sensor thread for each device object in the config. So every occurance of the thread only parses the information needed for that single device object. Therefor you don't need to create a linked list accross threads (which is also isn't threadsafe in this implementation). Just drop the linked list altogether and create an ID array. Loop over the ID's and use local variables (not arrays) for the remaining variables you now store in a linked list. Take the LM75 and LM76 as good examples for I2C reading and parsing.

If you updated the code in respect to my feedback, i will look further into it.

libs/protocols/bmp180.c
hstroh: Option `oversampling` sets the BMP180 hardware accuracy mode, selected by driver software via the variable `oversampling_setting`. This is only relevant for pressure readings (and not for temperature readings).

Mode | Parameter oversampling_setting | Internal no. of samples | Conversion time pressure max. [ms]
-----|--------------------------------|-------------------------|-----------------------------------
ultra low power | 0 | 1 | 4.5
standard | 1 | 2 | 7.5
high resolution | 2 | 4 | 13.5 
ultra high resolution | 3 | 8 | 25.5

hstroh:
Thanks for your feedback. Points 1 and 2 are clear but I have a comment and question regarding point 3. The code is based on protocol LM76. I2C reading and parsing is implemented in the same manner. Processing and storage of each sensor's calibration data makes the difference. In case of the BMP180 (as well as BMP085) the pressure and temperature data has to be compensated by the individual calibration data stored on each EEPROM. In order to avoid reading the calibration coefficients in each iteration step I decided to extend struct `bmp180data_t`(based on `lm76data_t`) by variables for storing the calibration values for each device specified via option `id` (having *multiple IDs* in mind). Disregarding my incorrect implementation (resulting from my lack of C knowledge), how could I store calibration data for each device id during initialization without using any kind of collection (list or array)? *Use local variables (not arrays)* implies that we are going to read calibration data for each device in each iteration step. Please comment.

CurlyMoo:
Reading the calibration data is perfect but just not in a global linked list. Just read those data in local values (when needed an array) before the loop starts.

hstroh:
Ok, fine. To store calibration data I'm now using an array of structure objects which is locally defined in function `bmp180Parse` as follows and initialized before we enter the loop:

	// structure for calibration data (stored in BMP180/085)
	struct {
		short ac1;
		short ac2;
		short ac3;
		unsigned short ac4;
		unsigned short ac5;
		unsigned short ac6;
		short b1;
		short b2;
		short mb;
		short mc;
		short md;
	} cd[bmp180data->nrid];

Before committing I've got another question regarding point 1. I could not find any documentation related to this feature. I've taken a look at protocols _Weather Underground_ and _Open Weather Map_ which currently are the only protocols that implement the update routine. A main difference is that these protocols use weather APIs and therefore use long-run poll intervals. Anyway, following their implementation I've added an option update (`u`) to the protocol. With respect to the default poll interval of 10 seconds (and short-term poll intervals for sensors in general) I've defined an update interval of 1 second. Now in the first run and each time the poll interval is reached a message with `"update": 1` and updated temperature and pressure values is broadcasted; according to the update interval a message with `"update": 0` and without temperature and pressure values is broadcasted in any other second. This leads to the web UI experience which I would expect: the update icon is highlighted for a second when data is updated. Does this meet your expectations?

CurlyMoo:
I didn't mean the `update` value. I mean this part:
https://github.com/pilight/pilight/blob/master/web/default/pilight.js#L873

That the webgui actually processes new values when they arrive.

The update button is meant not to overload website APIs. Also APIs have minimal update intervals themselves. They at least don't update their values more often than 10 to 15 minutes. So instead of just polling the API every 10 minutes, we just poll it once and let the user update the values when they want to. Also to relieve the sites. We at least don't allow updates more often than every 10 minutes.

hstroh:
Alright, please have another look.

CurlyMoo:
I have to admit when looking at the code again that you were right the first time. The initial struct implementation was valid. Sorry for that. I can merge the whole thing if you just remove the last two commits:

```
git push origin +beb6b14^bmp180
git push origin +7b9682e^:bmp180
git reset HEAD~2
```

hstroh:
At the end of the day anything was fine with the exception of the unnecessary help function!? Well, you should at least read and review committed code carefully before commenting it like I wrote my code carefully before creating a pull request. As you know your wrong advise has cost me some hours. Please keep that in mind in the future.

Now back to business: Even if the first implementation using global lists was "valid", why should we simply switch back to it? What's wrong with the current implementation? Isn't it valid as well? Isn't it better to use local variables to store calibration data as advised? Please tell me your thoughts so that I can learn a bit. 

You asked me to remove the last two commits, even though my last commit contained only one small change (check of a return value) which is surely relevant for any kind of implementation.

CurlyMoo:
> At the end of the day anything was fine with the exception of the unnecessary help function!? Well, you should at least read and review committed code carefully before commenting it like I wrote my code carefully before creating a pull request. As you know your wrong advise has cost me some hours. Please keep that in mind in the future.

I can assure you i did that and of course you could have told me my comment was incorrect. I'm open for feedback.

>  Isn't it better to use local variables to store calibration data as advised?

That's what i didn't read well enough. The data is already stored in a local scope. The global struct is just meant to define a new variable type which is thus used only locally.

> You asked me to remove the last two commits, even though my last commit contained only one small change (check of a return value) which is surely relevant for the robustness of any kind of implementation.

It's hard to see what the actual new value is in the last commit. Maybe you need to add it again. I was just helping you revert some changes in the quickest way.

hstroh:
Ok, so I will roll back as advised although I like the newer implementation much more in matters of readability and maintainability (no pointers, no (re)allocation of memory etc.). After that I will add the small change I mentioned (commit 7b9682e) again and will finally make another commit.

hstroh:
Here we go again. I've pushed another commit as discussed. After another review I've got one more question. The pointers for storing calibration data aren't freed at the moment. Do we need a block like the following for each pointer variable?

	if (bmp180data->ac1) {
		sfree((void *) &bmp180data->ac1);
	}

CurlyMoo:
Yes, what i would do is this:
```
for(i=0;i<size;i++) {
   sfree((void *)&bmp180data->ac1[i]);
}
sfree((void *)&bmp180data->ac1);
```
(not sure if size is the actual variable.

hstroh:
Your code corresponds to the freeing of pointer `id`. A difference between pointer `id` and the pointers for calibration data is that their "elements" are initialized in a different way like shown below. Could this be a reason for omitting the loop you proposed?

	bmp180data->id[bmp180data->nrid] = malloc(strlen(stmp) + 1);

	bmp180data->ac1[y] = (short) readReg16(bmp180data->fd[y], 0xAA);

CurlyMoo:
Partly because i always first free the array elements and secondly it's a reminder we are actually freeing an array for future developers.

hstroh:
Ok, in a few minutes I will push another commit with freeing code as suggested for each pointer storing calibration data.

hstroh:
Thanks for the hint! As I'm new to Valgrind I needed some time to get familiar with the verbose output it generates. I think that anything is fine although I might misinterpret the output. Here comes the summary: 

```
==23900== LEAK SUMMARY:
==23900==    definitely lost: 0 bytes in 0 blocks
==23900==    indirectly lost: 0 bytes in 0 blocks
==23900==      possibly lost: 1,496 bytes in 11 blocks
==23900==    still reachable: 348,641 bytes in 1,997 blocks
==23900==         suppressed: 0 bytes in 0 blocks
```

Most of the messages have one of the following forms and seem to be irrelevant:

```
==23900== Conditional jump or move depends on uninitialised value(s)
==23900==    at 0x48DB010: wiringXI2CReadReg8 (wiringX.c:237)
==23900==    by 0x490AC73: bmp180Parse (bmp180.c:166)
==23900==  Uninitialised value was created by a stack allocation
==23900==    at 0x49274D8: raspberrypiI2CReadReg8 (raspberrypi.c:709)

==23900== Conditional jump or move depends on uninitialised value(s)
==23900==    at 0x48DB19C: wiringXI2CReadReg16 (wiringX.c:257)
==23900==    by 0x490AC97: bmp180Parse (bmp180.c:70)
==23900==  Uninitialised value was created by a stack allocation
==23900==    at 0x4927494: raspberrypiI2CReadReg16 (raspberrypi.c:713)
```

I recognized another message which might be noteworthy:

```
==23900== Warning: noted but unhandled ioctl 0x720 with no size/direction hints
==23900==    This could cause spurious value errors to appear.
==23900==    See README_MISSING_SYSCALL_OR_IOCTL for guidance on writing a proper wrapper.
```

CurlyMoo:
Seems ok

hstroh:
Alright, finally I've added a check of the chip's version (1=BMP085, 2=BMP180) in addition to the check of the chip's id. I will push my commit now.

CurlyMoo:
Can you increase the webserver version number in `version.txt`? Then it's ready to commit.

hstroh:
Of course, should I increment it from 5.12 to 5.13?

CurlyMoo:
If 5.12 is the last, then indeed to 5.13

hstroh:
Done!

CurlyMoo:
Done

hstroh:
Thanks for merging and your support!

CurlyMoo:
Can you create the wiki page for the new protocol?

hstroh:
Yes, for sure. I will go for it tonight.

hstroh:
One more question, @CurlyMoo: The information provided on the protocol pages in the wiki (especially in sections config and settings) applies to the master version. As we both know the BMP180 protocol is currently only available in the dev version. How to deal with this? Should I add a page which provides config and settings information as implemented and also contains a note that says that any information on the page refers to the dev version only?

CurlyMoo:
Yes, describe it only applies to dev.

CurlyMoo:
Also, can you provide me the icon you used in at least 151x151 transparent PNG?

hstroh:
`pressure.png` is based on `humi.png` which has a size of 16x16 px in the version I used. Therefore I'm not able to provide an icon with a size greater than 16x16 px. If you tell me the source of `humi.png` I will create a new image based on a bigger version of `humi.png` (if available) and provide it as requested.

CurlyMoo:
https://github.com/pilight/illumina/blob/develop/app/src/main/res/drawable-xxhdpi/humidity_light.png

hstroh:
Thanks for the hint, I have analyzed the icons. To create a proper icon I have removed the raindrop, completed the circular line (by mirroring the left semicircle), auto cropped the image and scaled it up to a size of 113x113 px. Finally I have changed the size of the canvas to 151x151 px and centered the icon on it. To achieve lossless upscaling I have based my work on [illumina's SVG version of the icon](https://github.com/medien-dresden/illumina/blob/develop/resources/gfx/icons.svg). Here we go:
pressure-151x151px

CurlyMoo:
Thanks

hstroh:
The 16x16 px in the dev branch should be replaced by a downscaled version of the icon I've provided here. Should I create a pull request?

CurlyMoo:
I don't think it's noticeable right?

hstroh:
At a zoom level of 100% the only noticeable difference seems to be the color. Did you recolor the 16x16 px icons in any way? At a zoom level of 800% you can see that the circular line is more symmetric in the downscaled version. I would swap them (proper recoloring assumed). The current one is a bit hacky because lossy upscaling was involved in the creation process. It's up to you, please have a look.

Current version: pressure-current
Downscaled version: pressure-16x16px

CurlyMoo:
On my list :)

hstroh:
:+1:

hstroh:
I've added a [wiki page for the protocol](http://wiki.pilight.org/doku.php/bmp180) and a link to it on the [protocols page](http://wiki.pilight.org/doku.php/protocols).

CurlyMoo:
Look good!

PR #126, opened 2015-01-04
Protocol RSL366

kirichkov:
I'm resubmitting this pull request, along with updated code, because as it turns out the arctech_switch_old protocol uses different addressing of the remote devices, which makes it confusing what values to input for unit/id in the configuration file for RSL366 devices.


CurlyMoo:
Can you elaborate those differences?

kirichkov:
I've detailed them here - http://forum.pilight.org/Thread-RSL366T-Switches?pid=9164#pid9164

In short arctech uses groups 1 to 5 to identify Unit, while RSL uses 1 to 4; Arctech uses 6 to 10 to identify ID, while RSL uses 5 to 8. In RSL the position of the high pulse designates the Unit/ID. This is already properly documented in the wiki

CurlyMoo:
Can you also remove the rsl366 from `arctech_switch_old` and update it's version number in the same PR?

Thanks for the reminder.

kirichkov:
I saw no reference to rsl366 in arctech_switch_old.c, The only reference was the pulse length of 390, but that is present because of the Düwi Terminal Switches.
I've already amended the CMakeLists.txt as part of this PR.

libs/protocols/rsl366.c
```diff
@@ -0,0 +1,190 @@
...
+
+	rsl366->parseCode=&rsl366ParseCode;
+	rsl366->createCode=&rsl366CreateCode;
+	rsl366->printHelp=&rsl366PrintHelp;
+}
+
+#ifdef MODULE
+void compatibility(struct module_t *module) {
+	module->name = "rsl366";
+	module->version = "0.8";
```
CurlyMoo: If it's fully working, just make the version 1.0?

libs/protocols/rsl366.c
```diff
@@ -0,0 +1,190 @@
...
+	module->reqversion = "5.0";
+	module->reqcommit = NULL;
```
CurlyMoo: The required commit should be the same as in the other protocols...

kirichkov:
Done! Should be good to go!

PR #128, opened 2015-01-18
- Added languagesupport for DE

fl0eck:
- Updated jQuery mobile to 1.4.5
- Updated jQuery to 2.1.3


hstroh:
Hi fl0eck,

very nice. I've also created a branch with support for German language a few days ago (cp.  https://github.com/hstroh/pilight/commit/06b6a124cbc888d9707e7e98eee3c26549ceb51e) and I would love to discuss some translations with you. 

```
var language_de = {
		off: "Aus",
		on: "Ein",
		stopped: "Gestoppt",
		started: "Gestartet",
		toggling: "Schaltend",
		up: "+",
		down: "-",
		update: "Aktualisieren",
		available: "verfügbar",
		connecting: "Verbindung wird aufgebaut",
		connection_lost: "Verbindung verloren! Bitte klicken, um die Seite neu zu laden.",
		connection_failed: "Verbindungsaufbau fehlgeschlagen! Bitte klicken, um die Seite neu zu laden.",
		unexpected_error: "Unerwarteter Fehler"
	}
```

`toggling`, `connecting` and `loading` are present participles (verb + *ing*) and express active actions, i.e. actions that are currently going on and aren't finished yet. Consequently `toggling' has to be translated as *Schaltend* and `connecting` as *Verbindend* or *Verbindung wird aufgebaut*. Same applies to `loading`  (= *Ladend*) but this key has no usage at the moment. 

*Auf* and *Ab* are certainly correct translations of `up` and `down` but I've never seen a German remote control with these labels. From my point of view *+* and *-* are commonly used.

Last but not least an `unexpected_error` is not *unbekannt* but *unerwartet*.

One more point: In your translations of `connection_lost` and `connection_failed` you correctly use the nominalized verb *Laden* but together with the prefix *neu* which isn't involved in the nominalization it sounds/reads a bit unnatural to me. That's why I avoided nominalization in my version. Sadly there is no noun like *Neustart* which would solve the problem. Perhaps something like *zum erneuten Laden* would be better because it sounds more natural.

Please share your thoughts with me.

Cheers!

fl0eck:
Hiho hstroh,
thanks for your feedback. I'll just paste the 3 versions here, to compare them:

English version:
```
var language_en = {
	off: "Off",
	on: "On",
	stopped: "Stopped",
	started: "Started",
	toggling: "Toggling",
	up: "Up",
	down: "Down",
	update: "Update",
	loading: "Loading",
	available: "available",
	connecting: "Connecting",
	connection_lost: "Connection lost, touch to reload",
	connection_failed: "Failed to connect, touch to reload",
	unexpected_error: "An unexpected error occured"
}
```

Your version:
```
var language_de = {
        off: "Aus",
        on: "Ein",
        stopped: "Gestoppt",
        started: "Gestartet",
        toggling: "Schaltend",
        up: "+",
        down: "-",
        update: "Aktualisieren",
        available: "verfügbar",
        connecting: "Verbindung wird aufgebaut",
        connection_lost: "Verbindung verloren! Bitte klicken, um die Seite neu zu laden.",
        connection_failed: "Verbindungsaufbau fehlgeschlagen! Bitte klicken, um die Seite neu zu laden.",
        unexpected_error: "Unerwarteter Fehler"
    }
```
My version:
```
var language_de = {
	off: "Aus",
	on: "Ein",
	stopped: "Gestoppt",
	started: "Gestartet",
	toggling: "Schalten",
	up: "Auf",
	down: "Ab",
	update: "Aktualisieren",
	loading: "Laden",
	available: "Verfügbar",
	connecting: "Verbinde",
	connection_lost: "Verbindung verloren, berühre zum neu Laden",
	connection_failed: "Verbindung fehlgeschlagen, berühre zum neu Laden",
	unexpected_error: "Es ist ein unbekannter Fehler aufgetreten"
}
```
Lets start with `toggling`: The question is where `toggling` is found in the web-gui. Your translation is correct if you translate word by word. But I think `language.toggling` is used to "toggle 2 states". In german: `2 Zustände schalten` - If you say "the switch is toggling" the current state of the switch is "schaltend". The best thing would be to check the word in the gui, and find the better version.

Continue with `+` and `-`: I think your version of "up and down" is better. Lets use yours :smile: 

`loading` is missing in your version. I think we could use `Laden` of my version here.

I think at  `connection_lost` and `connection_failed` we shouldn't use `klicken` because the most web-gui sessions will run on mobile devices like smartphones or tablets. In the original version is also "touch" used. What to you say??

Last but not least the `unexpected_error`: You are right with `unexpected`->`unerwartet`, but I would pack this word in a sentence like in the original version.

Suggested optimized version:
```
var language_de = {
	off: "Aus",
	on: "Ein",
	stopped: "Gestoppt",
	started: "Gestartet",
	toggling: "Schalten",
	up: "+",
	down: "-",
	update: "Aktualisieren",
	loading: "Laden",
	available: "verfügbar",
	connecting: "Verbindung wird aufgebaut",
	connection_lost: "Verbindung verloren! Hier berühren um neu zu verbinden",
	connection_failed: "Verbindung fehlgeschlagen! Hier berühren um es erneut zu versuchen",
	unexpected_error: "Es ist ein unerwarteter Fehler aufgetreten"
}
```

Thanks for your input, and please let me hear your opinion for the "compined version of our 2 versions"

Cheers from austria!

hstroh:
Hi Flo,

thanks for your feedback and the comparison.

I've checked the code again. Corresponding to the wording the key `toggling` is only used in cases where the toggling process is pending and is exclusively used in conjunction with css class `ui-icon-loader`. Therefore *(Um)Schaltend* is the best choice from my point of view. Here comes an exemplary usage of the key:

```
		$('#'+sDevId+'_pendingsw').bind("click", function(event, ui) {
			event.stopPropagation();
			$('#'+sDevId+'_pendingsw').parent().removeClass('ui-icon-on').removeClass('ui-icon-off').addClass('ui-icon-loader');
			$('#'+sDevId+'_pendingsw').button('disable');
			$('#'+sDevId+'_pendingsw').text(language.toggling);
			// ...
```

In principle the same applies to key `loading` which should also be translated as an active action: *Ladend*. I didn't translate the key because it isn't used at the moment. Perhaps @CurlyMoo can tell us about the future of key `loading` so that we know if it should be included or dropped?

I'm fine with the use of *berühren* but we have to add a comma directly after it. The reason is that it is followed by an infinitive clause which starts with *um* (cp. [Kommaregeln, Erweiterter Infinitiv](http://de.wikipedia.org/wiki/Kommaregeln#Erweiterter_Infinitiv)).

I'm also fine with using a complete sentence for key `unexpected_error`. Because it's a sentence we should think about adding a punctuation mark at the end of it. The same applies to the sentences at the end of the translations of keys `connection_lost` and `connection_failed`. What's your point of view?

Regards,
Hendrik

PR #129, opened 2015-01-18
Fixed Bug in JS

fl0eck:
I pushed the false version with a little bug -_-


CurlyMoo:
Already fixed with the other merge.

PR #132, opened 2015-01-20
Corrected some mistakes in german translation

fl0eck:


hstroh:
Thanks for the updated translation, @fl0eck!

PR #133, opened 2015-01-20
protocol implementation for Quigg GT1000 RC with GT-FSI-08 switches

RinusW:
The implementation is compiled and tested with the latest master
on an RapberryPi and functions correctly
Committed are all files that has been changed by the build process


CurlyMoo:
Please do you're PR at the development branch.

PR #135, opened 2015-01-21
The Quigg Gt1000 protocol implementation adapted for compilation under t...

RinusW:
As far as I can see it is against the development branch now

latest development version

Signed-off-by: RinusW <rinus@vweert.nl>


CurlyMoo:
Do you know the difference between this protocol and the already existand quigg_switch?

wo-rasp:
As far as i know there are quite a few and i am willing to share my knowledge with all of you in order to provide even better support for the community.

CurlyMoo:
To be precise, maybe we should rename `quigg_switch` to something else.

RinusW:
That would not a bad idea. The older one as I have them is using 
remote GT-7000 with GT-7008AS switches. But I know that wo-rasp has a 
different kind but also in the 7000 range(??).
So, why not call it the quigg_gt7000 series. I called my own 
implementation quigg_7008 :-)

wo-rasp:
GT-7000 is the Productname of the Remote Control
Globaltronics ist the distributor here in Germany.
Devices are sold by Aldi, i am aware of the following brand names used: QUIGG, TEVION.

Devices known to me and and to be handled by the GT-7000 are:
GT-FSA-02 (switch) – (outdoor without dimmer)
GT-Fsi-04a (switch) – (switch without dimmer)
GT-Fsi-04d (switch) – (switch with integrated BRIGHT/DIMMER function)
GT-7008BS (switch)
GT-7008AD (switch with integrated BRIGHT/DIMMER function)

I propose to use GT as abbreviation of Globaltronics
Followed by  the functionality (_switch, _screen)
Followed by the model number oft he remote control (_1000, _7000)

Why _screen and not _dimmer:
The dim functionality of the devices supports brighter / dimmer using 15 levels, but the device does not support a dimlevel.

On the GT-7000 the dimm command is always send to the last device selected. Thus you need to turn a device ON or OFF before you can use the dimm functionality, this is a limitation of the GT-7000.
With pilight you do not have this limitation and you can select a device directly using the UP/DOWN  command from the GUI screen interface, in this case the device remembers its last dim level.

CurlyMoo:
Did we make a decision already to what we are going to rename the current `quigg_switch` to avoid confusion with the new `quigg_gt1000`

wo-rasp:
No, but what about:

quigg_gt7000 – switch functionality (GT-FSI-02, -04a, -04d, 7008-AS, 7008-BS, 7008-AD)

quigg_gt7000_di – dimmer (uses the screen elements oft he GUI to implement the dimmer functionality available with Button Row 5 of the GT-7000 remote control (used by the -04d, -AD devices)

 

Gesendet: Donnerstag, 29. Januar 2015 13:16
An: pilight/pilight
Cc: wo-rasp
Betreff: Re: [pilight] The Quigg Gt1000 protocol implementation adapted for compilation under t... (#135)

 

Did we make a decision already to what we are going to rename the current quigg_switch to avoid confusion with the new quigg_gt1000

RinusW:
I agree. Call  it quigg_gt7000.

CurlyMoo:
Done. What's the status of the wiki pages for these switches?

PR #136, opened 2015-01-25
Receive example - a way to send value to pilight-core

sirkubax:
This is an example for my question
http://forum.pilight.org/Thread-pilight-emulating-api-custom-daemon-as-a-data-input

The initial idea is: 
"Send/receive any receive code into core (config) of pilight"

The goal:
Connect non-pilight-compatible device with pilight-daemon.
In my case - I did want to send temperature values to the node defined in the system.
        "weather2": {
            "name": "Weather2",   
            "protocol": [ "alecto_wsd17" ],
            "id": [{
                "id": 4
            }],
            "temperature": 1469,  
        },

Here, I can connect to pilight-daemon (via any tcp socket)

{"message":"client controller"}
and send a "fake" receive message:
{"message":{"id":4,"temperature":1500},"origin":"receiver","protocol":"alecto_wsd17","uuid":"0000-00-00-46-99cc08"}

The same could be used to override device (switch) state without sending the signal
{"message":{"id":"B1","unit":34,"state":"on"},"protocol":"clarus_switch","origin":"receiver"}
(this would create change in system, without sending the code)

----------
Probably merge to current master is not a good idea :) but I would like to present this solution for further discussion.
With my current knowledge of the code I would consider implementation of this functionality via webserver Since it does support (conn->uri, "/send") and (conn->uri, "/config") it would be nice to have (conn->uri, "/receive"),  or could we allow sending value via (conn->uri, "/values")?

What is Your opinion?
It would be nice to build into api emulation of receive protocol.

==================
PS
Currently I can communicate with pilight daemon via 2-3 tcp sockets:
- client receiver to capture codes from gui or internall daemon
- client sender to send custom codes form my python-daemon
- client controller to send temperature value

The final goal it to write python-daemon as a controller (like gui) and communicate with daemon (via api?). Since python code is easy to manage it would create quick way for complicated tasks (in addition to "Rules" provided by pilight)

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -1077,7 +1077,7 @@ static void client_controller_parse_code(int i, JsonNode *json) {
             }
 
             char * pname;
-            json_find_string(json, "protocol", &pname)
+            json_find_string(json, "protocol", &pname);
 		    pilight.broadcast(pname, json);
         }
     }
@@ -1201,6 +1201,8 @@ static void socket_parse_data(int i, char *buffer) {
 		if(json_validate(buffer) == true) {
 #endif
 			json = json_decode(buffer);
+            char *output = json_stringify(json, NULL);
+            logprintf(LOG_DEBUG, "socket_parse_data valid json: %s", output);
 
 			/* The incognito mode is used by the daemon to emulate certain clients.
 			   Temporary change the client type from the node mode to the emulated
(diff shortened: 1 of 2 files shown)


CurlyMoo:
This is __really__ not what pull requests are for! Please discuss everything on the forum.

sirkubax:
"Probably merge to current master is not a good idea :) but I would like to present this solution for further discussion."

:)

Sure, let's discuss on the forum.

PR #140, opened 2015-02-06
Setting four

wo-rasp:
Some protocols most notably quigg_gt7000, arctech_*, brennenstuhl
do show symptoms of not working correctly with a send-repeat value of 10.
Based on real tests a send-repeat factor between 1 and 9 produce better results.
Thus the recommendation is to set this value as default to 4


CurlyMoo:
Again, why does this need 10 commits? Please pull first, then create your commit, and last push to a new branch.

PR #149, opened 2015-03-05
ping update

wo-rasp:


CurlyMoo:
Please read and follow this guide: https://github.com/pilight/pilight/blob/master/CONTRIBUTING.md
Also don't close this PR, all issues can be resolves until we got a nice mergeable piece :+1:

wo-rasp:
I have rebased my previous 2 commits plus the code cleanup commit into a single one but I do not know how to get rid of the merge commits from the development branch, however they should not cause any problems.

CurlyMoo:
Also, please make your commits a bit more verbose. What does it fix and maybe some background info.

wo-rasp:
Okay, a good method would be a link to the corresponding thread on the forum

http://forum.pilight.org/Thread-ping?pid=10444#pid10444

 

There were two problems:

1st problem was caused by the missing state condition (connected/disconnected) for the associated ip address of the thread

2nd problem was caused by the fact that the sender’s (e.q. pi’s) ip address was logged only, the remote ip address, triggering the error condition, was not logged.

 

 

Gesendet: Freitag, 6. März 2015 10:58
An: pilight/pilight
Cc: wo-rasp
Betreff: Re: [pilight] ping update (#149)

 

Also, please make your commits a bit more verbose. What does it fix and maybe some background info.

CurlyMoo:
I know, but it's good to be clear in our commits :smile:

PR #155, opened 2015-03-17
Sendmail action

NiekD:


libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+	struct JsonNode *jmessage = NULL;
+	struct JsonNode *jto = NULL;
+	struct JsonNode *jvalues = NULL;
+	struct JsonNode *jchild = NULL;
+	struct JsonNode *temp = NULL;
+	char *stmp = NULL;
+	int nrvalues = 0, itmp = 0;
+	
+	jsubject = json_find_member(arguments, "SUBJECT");
+	jmessage = json_find_member(arguments, "MESSAGE");
```
CurlyMoo: I don't see you check for the `SUBJECT` and `TO` pressence?

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+	} 
+	if(settings_find_string("mail-to", &stmp) != EXIT_SUCCESS) {
+		logprintf(LOG_ERR, "Sendmail: setting \"mail-to\" is missing in config");
+		return -1;
+	} 
+	if(settings_find_string("mail-from", &stmp) != EXIT_SUCCESS) {
+		logprintf(LOG_ERR, "Sendmail: setting \"mail-from\" is missing in config");
+		return -1;
+	} 
+	if(settings_find_string("mail-subject", &stmp) != EXIT_SUCCESS) {
```
CurlyMoo: The `SUBJECT` is also part of the mail action arguments, so why do we have a `mail-subject` setting?

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+	} 
+	if(settings_find_string("smtp-user", &stmp) != EXIT_SUCCESS) {
+		logprintf(LOG_ERR, "Sendmail: setting \"smtp-user\" is missing in config");
+		return -1;
+	} 
+	if(settings_find_string("smtp-password", &stmp) != EXIT_SUCCESS) {
+		logprintf(LOG_ERR, "Sendmail: setting \"smtp-password\" is missing in config");
+		return -1;
+	} 
+	if(settings_find_string("mail-to", &stmp) != EXIT_SUCCESS) {
```
CurlyMoo: See comment below this one.

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+	jsubject = json_find_member(arguments, "SUBJECT");
+	jmessage = json_find_member(arguments, "MESSAGE");
+	jto = json_find_member(arguments, "TO");
+	
+	if(jmessage != NULL) {	
+		if((jvalues1 = json_find_member(jmessage, "value")) != NULL) {
+			if((jval1 = json_find_element(jvalues1, 0)) != NULL && jval1->tag == JSON_STRING){
+				message = jval1->string_;
+			}
+			subject = ssubject;
```
CurlyMoo: Why don't you use `subject` directly on line 143?

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+			if(jsubject != NULL) {	
+				if((jvalues2 = json_find_member(jsubject, "value")) != NULL) {
+					if((jval2 = json_find_element(jvalues2, 0)) != NULL && jval2->tag == JSON_STRING){
```
CurlyMoo: Can you add spaces between the `hook` and the `accolade`?

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+			mail.subject = MALLOC(strlen(subject)+1);
+			mail.message = MALLOC(strlen(message)+1);
+
+			strcpy(mail.from, sfrom);
+			strcpy(mail.to, to);
+			strcpy(mail.subject, subject);
+			strcpy(mail.message, message);
+			
+			if(sendmail(shost, suser, spassword, sport, &mail) != 0) {
+				logprintf(LOG_DEBUG, "Sendmail failed with message \"%s\"", message);
```
CurlyMoo: I assume the message here is some sort of error message, but it isn't. So why the error like this?

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+			FREE(mail.from);
+			FREE(mail.to);
+			FREE(mail.subject);
+			FREE(mail.message);
+		}
+	}
+	return 0;
+}
+
+#ifndef MODULE
```
CurlyMoo: Add the windows define as well.

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+
+	options_add(&action_sendmail->options, 'a', "SUBJECT", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_STRING, NULL, NULL);
+	options_add(&action_sendmail->options, 'b', "MESSAGE", OPTION_HAS_VALUE, DEVICES_VALUE, JSON_STRING, NULL, NULL);
+	options_add(&action_sendmail->options, 'c', "TO", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_STRING, NULL, NULL);
+
+	action_sendmail->run = &actionSendmailRun;
+	action_sendmail->checkArguments = &actionSendmailArguments;
+}
+
+#ifdef MODULE
```
CurlyMoo: Windows defines?

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+void compatibility(struct module_t *module) {
+	module->name = "sendmail";
+	module->version = "1.0";
+	module->reqversion = "6.0";
+	module->reqcommit = NULL;
```
CurlyMoo: Please check which commit included the sendmail library.

libs/actions/sendmail.c
NiekD: Correct,  these arguments are optional and if omitted the corresponding 
setting will be used.



CurlyMoo schreef op 17-3-2015 om 19:01:

libs/actions/sendmail.c
NiekD: The mail-subject setting is the default if the SUBJECR argument is omitted.

CurlyMoo schreef op 17-3-2015 om 19:02:

libs/actions/sendmail.c
NiekD: See previous reply.

CurlyMoo schreef op 17-3-2015 om 19:03:

libs/actions/sendmail.c
CurlyMoo: I'm very enthousiastic about having these settings in the settings.

libs/actions/sendmail.c
NiekD: Good question.


CurlyMoo schreef op 17-3-2015 om 19:04:

libs/actions/sendmail.c
NiekD: Sure.

CurlyMoo schreef op 17-3-2015 om 19:04:

libs/actions/sendmail.c
NiekD: sendmail is in the mail library and returns just 0 (success) or -1 (error)

CurlyMoo schreef op 17-3-2015 om 19:06:

CurlyMoo: I know, but you text: `Sendmail failed with message \"%s\"` tells me that i will read the error message from the sendmail library, but i actually will see my own message.

libs/actions/sendmail.c
NiekD: I know nothing about the Windows part yet:-(

CurlyMoo schreef op 17-3-2015 om 19:06:

libs/actions/sendmail.c
NiekD: See previous reply

CurlyMoo schreef op 17-3-2015 om 19:06:

libs/actions/sendmail.c
CurlyMoo: https://github.com/pilight/pilight/blob/development/libs/actions/switch.c#L152

libs/actions/sendmail.c
NiekD: OK

CurlyMoo schreef op 17-3-2015 om 19:07:

libs/actions/sendmail.c
NiekD: What do you mean here?


CurlyMoo schreef op 17-3-2015 om 20:46:

CurlyMoo: I was missing the **not** :)

libs/actions/sendmail.c
NiekD: Correct, you see the message that couldn't be sent.

CurlyMoo schreef op 17-3-2015 om 20:54:

libs/actions/sendmail.c
NiekD: OK, I see
CurlyMoo schreef op 17-3-2015 om 20:56:

libs/actions/sendmail.c
```diff
@@ -0,0 +1,226 @@
...
+				logprintf(LOG_DEBUG, "Sendmail failed with message \"%s\"", message);
+			}
+			
+			FREE(mail.from);
+			FREE(mail.to);
+			FREE(mail.subject);
+			FREE(mail.message);
+		}
+	}
+	return 0;
```
CurlyMoo: This function always succeeds despite the message being actually sent or not...

libs/actions/sendmail.c
NiekD: Now I understand, but I do find them usefull.

In my own case, all messages sent from rules must go to the same 
recipient (myself), with always  the same sender (pilight) and most of 
them with the same subject (pilight event). In those rules I only need 
to use the "MESSAGE" argument , because the others are all in the settings.

But if you don't like these setting let's throw them out.

CurlyMoo schreef op 17-3-2015 om 20:58:

CurlyMoo: :+1: I think you're the exception here.

libs/actions/sendmail.c
NiekD: Correct, in fact this is the same as with the pushover action  (which I 
took as the start for the sendmail action btw), but I should have used 
LOG_ERR in the logprintf instead of LOG_DEBUG I think.

CurlyMoo schreef op 17-3-2015 om 21:24:

libs/actions/sendmail.c
NiekD: I guess that I'm the only one using it until now, so how can I be the 
exception? :-) :-) :-)

But no kidding, do you wish to drop all 3 mail-xxxx settings?


CurlyMoo schreef op 17-3-2015 om 21:29:

CurlyMoo: We need the `mail-from` somewhere. Maybe call it `smtp-sender`?

NiekD: Why not.

So  will make TO and SUBJECT mandatory arguments and only the sender 
addres will come from a mandatory smtp-sender setting (so no FROM 
argument), right ?

CurlyMoo schreef op 17-3-2015 om 21:53:

CurlyMoo: Nope, i think the from can be global,

libs/actions/sendmail.c
NiekD: I know you introduced the mail library in 6e2e41d and tried to get the 
commit number the way you indicated, but I get this:

root@raspi2:/home/pi/pilight# git checkout 6e2e41d
HEAD is now at 6e2e41d... Various fixes
root@raspi2:/home/pi/pilight# git describe
fatal: No tags can describe '6e2e41dd7bc241fc48e1f4a638da173d71789cc3'.

I suppose the reqcommit must be set to a number that should be 
represented by a tag?

CurlyMoo schreef op 17-3-2015 om 19:07:

CurlyMoo:
```
# git describe 6e2e41d
v6.0-51-g6e2e41d
```
The number is 51.

NiekD: Hmmm, if I do the same I still get a fatal error.

git describe 6e2e41d
fatal: No tags can describe '6e2e41dd7bc241fc48e1f4a638da173d71789cc3'.

Don't know what's wrong, but anyway, I understand it is 51 and I will 
put that in.

CurlyMoo schreef op 18-3-2015 om 17:42:
> |git describe 6e2e41d|

libs/actions/sendmail.c
CurlyMoo: Indeed. leave it return 0

PR #158, opened 2015-03-21
allow underscore and hyphen in device names

martinr63:


CurlyMoo:
Tested locally?

martinr63:
yes, but without rules

CurlyMoo:
Can you test with rules?

martinr63:
Tested with rules, everything okay

CurlyMoo:
Thanks

PR #163, opened 2015-03-29
added smtp settings for sendmail action

NiekD:
added smtp settings with EVENTS define
improved regex for e-mail addresses


libs/pilight/config/settings.c
```diff
@@ -458,6 +458,90 @@ static int settings_parse(JsonNode *root) {
...
+				logprintf(LOG_ERR, "config setting \"%s\" must contain an smtp host address in the format of \"smtp.gmail.com\"", jsettings->key);
+				have_error = 1;
+				goto clear;
+			} else if(jsettings->string_ == NULL) {
+				logprintf(LOG_ERR, "config setting \"%s\" must contain an smtp host address in the format of \"smtp.gmail.com\"", jsettings->key);
+				have_error = 1;
+				goto clear;
+			} else if(strlen(jsettings->string_) > 0) {
+#if !defined(__FreeBSD__) && !defined(_WIN32)
+				char validate[] = "^(smtp)\\.[a-zA-Z0-9_]+\\.[a-zA-Z]{2,3}$";
```
CurlyMoo: I'm not sure this is a valid regex: `smtp.blabla.co.uk`.

libs/pilight/config/settings.c
```diff
@@ -458,6 +458,90 @@ static int settings_parse(JsonNode *root) {
...
+				reti = regcomp(&regex, validate, REG_EXTENDED);
+				if(reti) {
+					logprintf(LOG_ERR, "could not compile regex for %s", jsettings->key);
+					have_error = 1;
+					goto clear;
+				}
+				reti = regexec(&regex, jsettings->string_, 0, NULL, 0);
+				if(reti == REG_NOMATCH) {logprintf(LOG_ERR, "No Match: %s", jsettings->string_);}
+				if(reti == REG_NOMATCH || reti != 0) {
```
CurlyMoo: Please check this condition when compared to the one just above.

libs/pilight/config/settings.c
NiekD: You are right. I will change the regex so it will accept such url's too

CurlyMoo schreef op 29-3-2015 om 12:45:

libs/pilight/config/settings.c
NiekD: Forgot to delete that first one

CurlyMoo schreef op 29-3-2015 om 12:46:

libs/pilight/config/settings.c
NiekD: Did some searching for smtp server adresses and came to the conclusion 
that the current regex is much to strict.

Examples:

smtp.x.y
mail.x.y
mailhost.x.y
relay.x.y
post.x.y
mail.username.x.y
mail.domainname.x.y

These are just Dutch smtp hosts, so there is a big chance that there 
will be more.
So, if you agree,  I will change the regex to:

^([a-zA-Z0-9\\_\\-]){2,20}(\\.([a-zA-Z0-9\\_\\-]){2,20}){2,3}$

This will allow any prefix ,subdomains, and TLD's with more than 3 
characters (eg .amsterdam) and will still catch the most common typo's.

CurlyMoo schreef op 29-3-2015 om 12:45:

CurlyMoo: Agree

PR #176, opened 2015-04-27
Raw printout optimized

wo-rasp:
The current version of pilight-raw prints the hardware id before each pulse and inserts a CR after each pulse. This version will insert a CR only if the generic footer condition is true, e.q. a pulse with a duration > 5100 is found, and will add the hardware id after the CR.
This will greatly simplify recognition of pulsetrains and allow easy copy of pulsetrains to pilight-send with the raw option.
The drawback may be on machines with multiple hardware boards and on a distributed ADHOC network, receiving data at the same time, this scenario is not tested, and not each single pulse is prefixed with the hardware id, thus perhaps this version should be made available as pilight-rawf (f for formatted).


wo-rasp:
Accidentally requested it against master ----

PR #177, opened 2015-04-27
Modified raw module with structured printout

wo-rasp:
The current version of pilight-raw prints the hardware id before each pulse and inserts a CR after each pulse. This version will insert a CR only if the generic footer condition is true, e.q. a pulse with a duration > 5100 is found, and will add the hardware id after the CR.
This will greatly simplify recognition of pulsetrains and allow easy copy of pulsetrains to pilight-send with the raw option.
The drawback may be on machines with multiple hardware boards and on a distributed ADHOC network, receiving data at the same time, this scenario is not tested, and not each single pulse is prefixed with the hardware id, thus perhaps this version should be made available as pilight-rawf (f for formatted).

### DIFF
--- a/raw.c
+++ b/raw.c
@@ -89,13 +89,19 @@ int main_gc(void) {
 }
 
 void *receiveOOK(void *param) {
-	int duration = 0;
+	int duration = 0, iLoop = 0;
 
 	struct hardware_t *hw = (hardware_t *)param;
 	while(main_loop && hw->receiveOOK) {
 		duration = hw->receiveOOK();
+		iLoop++;
 		if(duration > 0) {
-			printf("%s: %d\n", hw->id, duration);
+			if (duration > 5100) {
+				printf(" %d -#: %d\n%s: ",duration, iLoop, hw->id);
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Isn't this exactly what pilight-debug does?

wo-rasp:
No, it is not.

pilight-debug is trying to catch a footer pulse and then starts recording.

This may fail and if we have unusual pulse durations, pilight-debug may report nothing.

 

Like the original pilight-raw, all incoming data is printed out.

However a CR is only printed if a pulse > 5100 is detected and before the CR the number of pulses received is printed.

You will be able to spot pulsetrains more or less immediately, easily detect “garbeled data”, noise pulses, pauses etc.

 

Gesendet: Montag, 27. April 2015 17:48
An: pilight/pilight
Cc: wo-rasp
Betreff: Re: [pilight] Modified raw module with structured printout (#177)

 

Isn't this exactly what pilight-debug does?

CurlyMoo:
Can you make this a parameter to pilight-raw?

PR #181, opened 2015-06-19
fixed bug in x.10 protocol

NiekD:
This bug caused sending of the same raw codes for unit codes with a numeric part of 8 and 9.


CurlyMoo:
I knew you could do it :smile:

PR #184, opened 2015-07-11
Edak added dogcollar protocol

edak:


libs/pilight/protocols/433.92/dogcollar.c
```diff
@@ -0,0 +1,184 @@
...
+        createFunction(function);
+        createFooter();
+		dogcollar->rawlen = RAW_LENGTH;
+	}
+	return EXIT_SUCCESS;
+}
+
+static void printHelp(void) {
+        printf("\t -i --id=id\t\tcontrol a device with this id (1-29999)\n");
+        printf("\t -f --function=function\t\t (1=light, 2=beep, 3=vibrate-low, 4=vibrate-high)\n");
```
CurlyMoo: I would like to change this into e.g. this:
```
\t -l --light\t\t...
\t -b --beep\t\t...
\t -f --vibratelow\t\t...
\t -t --vibratehigh\t\t...
```

Internally you can still use the same logic ofc.

CurlyMoo:
Can you please check this guide and fix the PR (not reopen!) accordingly:
https://github.com/pilight/pilight/blob/master/CONTRIBUTING.md

Please also fix the alinements some tabs are actually spaces.

libs/pilight/protocols/433.92/dogcollar.c
edak: Hi CurlyMoo,
There are no controls for the GUI that would support this change, hence why I had to go with a numeric value. If you can point me in the direction otherwise then I will be happy to re-design this protocol structure as you see fit. The only options of type JSON_STRING are states which have a text value (upon button press) returned of either "down" or "up"

CurlyMoo: > There are no controls for the GUI that would support this change, hence why I had to go with a numeric value.

Correct, but i'm planning a certain logic to allow this in the future. The change helps with it.

edak:
fixed the alignments and tabs

CurlyMoo:
Can you also do everythign as described in the link i posted earlier.

woutput:
I think @CurlyMoo hints at checklist number 7: *Make sure your pull-request contains one single commit.* My post in [this thread] (http://forum.pilight.org/Thread-How-to-combine-commits-when-using-github-online) might help you.

I have another question:
You now have called this protocol "dog collar protocol". I think this is a rather generic name; could you maybe add the brand and/or type in the protocol name? For example "Pet lovers - bark-stop dog collar protocol". This would allow other dog collar protocols of other dog collar brands and/or types to be added in the future.

wo-rasp:
Do we want to add that protocol ? If yes i would volunteer to collapse the work into one single commit.

PR #194, opened 2015-09-04
Fix -v option of pilight-control

pilino1234:
Arguments are now separated by "=" and not by ",=" anymore


CurlyMoo:
The komma is necessary because you need to be able to do:
`-v "temperature=1,humidity=2"`

PR #196, opened 2015-09-19
Created ISNOT operator

NiekD:
This operator takes two string operands (just like the IS operator) and
evaluates to TRUE if the operands are NOT equal.

### DIFF
--- a/libs/pilight/events/operators/isnot.c
+++ b/libs/pilight/events/operators/isnot.c
@@ -1,0 +1,55 @@
+/*
+	Copyright (C) 2013 - 2014 CurlyMo & Niek
+
+	This file is part of pilight.
+
+	pilight is free software: you can redistribute it and/or modify it under the
+	terms of the GNU General Public License as published by the Free Software
+	Foundation, either version 3 of the License, or (at your option) any later
+	version.
+
+	pilight is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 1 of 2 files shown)


CurlyMoo:
Isn't this possible in another way?

NiekD:
Op 19-9-2015 om 22:22 schreef CurlyMoo:

wo-rasp:
I t is beneficial to have this operator.
Example: On s_mfy, we have the 3 button interface and three states: Down, My ( Stopped), Up.
If i want to implement a rule to move the screen down, i currently need to test for MY and UP, if i want to move it down, with ISNOT, i simply can ask with one stetement ifnot(DOWN) -move DOWN ... even easy to understand.

NiekD:
Ofc I agree ;). 
I myself am using the COLOR option of generic_label devices as variables.  If such a "variable" can have more than 2 different values, I am using the ISNOT operator to check for the absence of a specific value. If I would only have the IS operator I would have to check for the presence of any value the "variable" might have. 
Simple example: values can represent the conditions CLOSED, OPEN and LONG OPEN. With IFNOT I can test if the condition is not CLOSED, otherwise I would need to check for both OPEN and LONG OPEN.

wo-rasp:
The ISNOT operator is of more use than the the IS operator.
If I have two different states, IS and ISNOT are equivalent.
If I have three different states, I can use one ISNOT operator to rule out the two other conditions.
I propose to get the operator integrated.

PR #198, opened 2015-10-06
Allow any pin number for parameter gpio

woutput:
I don't know what I'm doing, I haven't tested it and I just copied this part from https://github.com/pilight/pilight/blob/master/libs/pilight/protocols/GPIO/relay.c
So please verify and/or test thoroughly :)
This hopefully fixes http://forum.pilight.org/Thread-Bug-report-gpio-switch-does-not-accept-high-pin-numbers


woutput:
Is this some kind of regular expression?

pilino1234:
You should make PRs against the development branch, not the master (which has the latest stable version)

woutput:
Oops, sorry, didn't know that.
Should I try again on that branch?

pilino1234:
Yes, you'll need to base your changes on the developmental code

woutput:
OK, thanks for the help.

PR #202, opened 2015-11-01
Typo's

woutput:
Corrected typo's and hyperlink.

In addition, I don't understand the rule "no unnecessary spaces".
The example `if ( 1 == 1 )` changes to `if(1 == 1)` but then why not to `if(1==1)` since I believe the spaces around `==` are also unnecessary. If the intention is to come to `if(1 == 1)` then I think more detailed rules are needed.
In addition, this rule appears to be unclear and/or incorrectly implemented. /master/debug.c contains for example `x=(double)i/pulselen;` (line 57) and `main_loop = 0;` (line 64) so with and without spaces around the `=`-sign.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@ Rules for contributing to pilight
 First of all, thanks for making this project better. Without your help, pilight wouldn't have come to what it is now!
 
 **BUT**, be aware that contributing to a big project shouldn't be considered a playground. It also isn't meant for you to learn `git`.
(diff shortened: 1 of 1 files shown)


CurlyMoo:
> The example if ( 1 == 1 ) changes to if(1 == 1) but then why not to if(1==1) since I believe the spaces around == are also unnecessary. If the intention is to come to if(1 == 1) then I think more detailed rules are needed.

True. I mean around the hooks.

> In addition, this rule appears to be unclear and/or incorrectly implemented. /master/debug.c contains for example x=(double)i/pulselen; (line 57) and main_loop = 0; (line 64) so with and without spaces around the =-sign.

I'm not perfect either :smile:

PR #204, opened 2015-11-06
radxa rock board added emulate both edge triggered interrupts

jack-ma:
Hello, we have added and tested both edge triggered interrupts function for radxa rock board.
https://github.com/radxa/linux-rockchip/commit/6650a736ff6a9d8043e8754d416f70d3d83ca051


CurlyMoo:
Thanks, but i will sync with wiringX myself.

PR #205, opened 2015-11-11
Added Eurodomest switch protocol

woutput:
Added Eurodomest switch protocol and removed old "support" from arctech_switch_old.
This is my first protocol implementation, so please review carefully.
It is implementing the protocol requested in [this thread] (http://forum.pilight.org/Thread-Eurodomest), that is, the protocol that fits most Eurodomest switches, but not all.
I have tested it, it is able to receive the remote control signals and send own signals. The GUI buttons are synchronized with the remote control.


libs/pilight/protocols/433.92/eurodomest_switch.c
```diff
@@ -0,0 +1,336 @@
...
+#include "../../core/log.h"
+#include "../protocol.h"
+#include "../../core/binary.h"
+#include "../../core/gc.h"
+#include "eurodomest_switch.h"
+
+#define PULSE_MULTIPLIER	3
+#define MIN_PULSE_LENGTH	142
+#define MAX_PULSE_LENGTH	426
+#define AVG_PULSE_LENGTH	284
```
CurlyMoo: Are these pulses this inaccurate?

libs/pilight/protocols/433.92/eurodomest_switch.c
```diff
@@ -0,0 +1,336 @@
...
+		json_append_member(eurodomest_switch->message, "unit", json_mknumber(unit, 0));
+	}
+
+	if(state == 1) {
+		json_append_member(eurodomest_switch->message, "state", json_mkstring("on"));
+	} else {
+		json_append_member(eurodomest_switch->message, "state", json_mkstring("off"));
+	}
+
+	if(learn == 1) {
```
CurlyMoo: Do these switches actually support learning?

libs/pilight/protocols/433.92/eurodomest_switch.c
```diff
@@ -0,0 +1,336 @@
...
+		}
+	}
+
+	// The last 4 bits contain the unit, state and all information
+
+	int unit = 0;
+	int all = 0;
+	int state = 0;
+
+	if(binary[20] == 0 && binary[21] == 0 && binary[22] == 0 && binary[23] == 0) {
```
CurlyMoo: This can be done a bit better. E.g.
```
x = binToDecRev(binary, 20, 23);

switch(x) {
	case 0:
		unit = 1;
		all = 0;
		state = 1; // on
	break;
	case 1:
		unit = 1;
		all = 0;
		state = 0; // on		
	break;
	case 2:
		unit = 2;
		all = 0;
		state = 0; // off	
	break;
	...
}
```

woutput: Thank you for the suggestion. I was indeed looking for something better here, but don't know it.
I don't agree with your proposal though, because it may be shorter, but it is also harder to understand. The function of this code block is to act on received *bits*, not just "a value".
I was actually looking for something like `if (binary[20...23] == "0110")` Do you know anything like that in C?

libs/pilight/protocols/433.92/eurodomest_switch.c
woutput: Yes they do, I haven't explicitly tested this function though. I will.

libs/pilight/protocols/433.92/eurodomest_switch.c
CurlyMoo: Yes, that's what i proposed. But instead of comparing binary you compare decimal.

libs/pilight/protocols/433.92/eurodomest_switch.c
woutput: Well, yes and no:
Yes: see the graphs in the second sheet of [this attachment] (http://forum.pilight.org/attachment.php?aid=595) from [the referred thread] (http://forum.pilight.org/Thread-Eurodomest)
No: but I suppose there is no disadvantage in accepting a wider range? Therefore I thought: the wider, the more robust.

CurlyMoo: The wider the range the more false positives.

libs/pilight/protocols/433.92/eurodomest_switch.c
woutput: I understand your suggestion, but in my opinion it's not better than my current implementation, *"because it may be shorter, but it is also harder to understand. The function of this code block is to act on received **bits**, not just "a (decimal) value (without meaning)"."*
(Why) do you think using decimal would be better?

libs/pilight/protocols/433.92/eurodomest_switch.c
woutput: Ah OK, thank you.
Let me reconsider the timing...

woutput:
OK, updated.

Some questions:
1. Should I merge the commits into one before we continue discussion? Or only just before the merge to pilight/development?
2. From the [protocol development page] (http://www.pilight.org/modules/protocols/) I understand that I should use parseRaw instead of parseCode, because this code looks at the timing of the raw pulses. Is there a way to use the ->code[] from parseCode? Then how can I supply the timings of the pulses? Or should I just keep the current code and rename parseCode to parseRaw?
3. Learning a switch works, even when not using -l (works for both -t and -f), so using 10 repeats is sufficient, 40 is not needed. Should I remove the learn part? Should I try to reduce the number of repeats for send and/or learn?
4. Learning only accepts IDs 524288 (2^19) up to 1048575 (2^20-1), not lower IDs. Is this acceptable for now or should I continue investigation?

Thanks again for reviewing.

CurlyMoo:
> 1. Should I merge the commits into one before we continue discussion? Or only just before the merge to pilight/development?

When final.

> 2. From the protocol development page I understand that I should use parseRaw instead of parseCode, because this code looks at the timing of the raw pulses. Is there a way to use the ->code[] from parseCode? Then how can I supply the timings of the pulses? Or should I just keep the current code and rename parseCode to parseRaw?

parseRaw, parseBinary and parseCode are obsolete. It's now only parseCode.

> 3. Learning a switch works, even when not using -l (works for both -t and -f), so using 10 repeats is sufficient, 40 is not needed. Should I remove the learn part? Should I try to reduce the number of repeats for send and/or learn?

This can happen, but multiple repeats are better.

> 4. Learning only accepts IDs 524288 (2^19) up to 1048575 (2^20-1), not lower IDs. Is this acceptable for now or should I continue investigation?

As long as you notify the user when using a wrong ID in learn mode.

> 5. How should I divide the validating/parsing functionality between static int validate(void) and static void parseCode(void)? Should I move everything for validation to validate? Or keep only a little (as now)? I guess this is a trade-off between speed and selectivity?

As it is. Indeed a trade-off.

woutput:
Thanks, now working on
*Learning only accepts IDs 524288 (2^19) up to 1048575 (2^20-1), not lower IDs. Is this acceptable for now or should I continue investigation?*

*As long as you notify the user when using a wrong ID in learn mode.*

woutput:
There is one thing about the ID though. When I was testing my switch to learn different IDs I found that the MSB should always be 1. But when analyzing the IDs of the posted remote controls, I found multiple where the MSB is 0. So I suggest to keep all 20 bits as ID, but not all IDs will work in all systems. I think this is acceptable, but what do you think?
If OK then I think I'm done for now. Please review (again).

CurlyMoo:
> So I suggest to keep all 20 bits as ID, but not all IDs will work in all systems. I think this is acceptable, but what do you think?

Ok

CurlyMoo:
> Please review (again).

Seems ok

woutput:
Tnx

woutput:
What do you think? OK to merge to pilight/development?
Then I will merge into one commit.

CurlyMoo:
I'm waiting for confirmation of other users.

woutput:
Hi CurlyMoo, it has been some time, but someone has tested my protocol implementation and it works successfully, see https://forum.pilight.org/Thread-Eurodomest?pid=17826#pid17826
Is this sufficient to accept in dev?
PS Sorry for bad formatting, but this is from smartphone

CurlyMoo:
Seems to be fine. As long as you follow the contribution rules.

woutput:
> Seems to be fine

Thank you

> As long as you follow the contribution rules

Did you mean to say "Please merge into one commit" then we agreed:
> Should I merge the commits into one before we continue discussion? Or only just before the merge to pilight/development?
>> When final.

so I'll merge now.
If there's something else, please let me know :smile:

PR #214, opened 2015-12-07
Fix RSL366T remote code parsing

dominikkarall:
Fixed RSL366T remote code parsing
Reduced the number of false positives for arctech_switch_old based on the description (https://wiki.pilight.org/doku.php/arctech_switch_old) which says bit[11]=0

### DIFF
--- a/libs/pilight/protocols/433.92/rsl366.c
+++ b/libs/pilight/protocols/433.92/rsl366.c
@@ -31,15 +31,18 @@
 #include "rsl366.h"
 
 #define PULSE_MULTIPLIER	3
-#define MIN_PULSE_LENGTH	385
+#define MIN_PULSE_LENGTH	375
 #define MAX_PULSE_LENGTH	395
-#define AVG_PULSE_LENGTH	390
-#define RAW_LENGTH				50
+#define AVG_PULSE_LENGTH	480
+#define RAW_LENGTH		50
 
 static int validate(void) {
 	if(rsl366->rawlen == RAW_LENGTH) {
 		if(rsl366->raw[rsl366->rawlen-1] >= (MIN_PULSE_LENGTH*PULSE_DIV) &&
-		   rsl366->raw[rsl366->rawlen-1] <= (MAX_PULSE_LENGTH*PULSE_DIV)) {
+                   rsl366->raw[rsl366->rawlen-1] <= (MAX_PULSE_LENGTH*PULSE_DIV) &&
+                   rsl366->raw[rsl366->rawlen-3] <= (AVG_PULSE_LENGTH*PULSE_MULTIPLIER) &&
+                   rsl366->raw[rsl366->rawlen-7] <= (AVG_PULSE_LENGTH*PULSE_MULTIPLIER) &&
+                   rsl366->raw[rsl366->rawlen-11] <= (AVG_PULSE_LENGTH*PULSE_MULTIPLIER)) {
 			return 0;
 		}
 	}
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Pleaes do your PR's according to the contribution guide.

CurlyMoo:
Also, don't close it!

dominikkarall:
@CurlyMoo how can I update the pull request? I would like to do a pull request from here:
https://github.com/dominikkarall/pilight/tree/rsl366tfix2

CurlyMoo:
Don't redo it, but fix the current one. Information can be found in the contribution guide.

dominikkarall:
I created a new branch as mentioned in the contribution guide and also created a single commit there. I'm just not sure if that one would match this PR if I do a pull from the rsl366fix2 branch.

CurlyMoo:
As i said, don't create new branches, but fix the current one. A link on how to do this can be found in the documentation.

woutput:
I think it would be very strange if the remote control actually sent sometimes 49 transitions and sometimes 50. I really think that the 49 is actually caused by distortion (as I expect it also to be the case in the Eurodomest protocol that I created).
Can you explain why you think this would be valid behaviour?


Does this protocol only allow receiving? Is there no sending part? (most protocols I have seen do both)


Average pulse length is larger than maximum (and minimum) pulse length. This seems contradicting. Can you explain?


Thank you,
Wouter

dominikkarall:
I think you are right, the average pulse length can remain at 390. I changed it because of following pulses I received:
````
1364 417 1102 505
450 1230 1175 515 
309 1447 1106 481 
355 1406 261 1399 
389 1249 277 1419 
273 1456 1162 625 
170 1315 1402 429 
224 1325 1202 416 
388 1402 1144 599 
198 1365 1407 487 
148 1347 1148 450 
529 1389 1000 609 
182 13269 -#: 50
````

Due to the fact that I was checking the 4th pulse instead of the 3rd one, I increased it to 480. Otherwise a 4th pulse >585 would be treated as low instead of high (e.g.609). As I'm checking the 3rd pulse again, I will revert that one and test it.

Regarding the number of pulses. Could it be that the low pass filter truncates one pulse? My remote sends more often 49 pulses instead of 50 pulses if I look at the pilight-raw -L output.

dominikkarall:
Ok, I can remove pulse length 49 and it still works. Really strange that pilight-raw -L shows me mainly 49 instead of 50. I'm going to revert that one.

I tried to change the average pulse length back to 390, but I have to say, that it works much worse. Just a few signals are properly decoded. Using 480 works great.

I also identified that the min pulse length seems to be incorrect. There are some pulses I receive with 380. If I change the min pulse length to something lower 380 those signals get also properly decoded.

dominikkarall:
I just tested my Brennenstuhl RCS1000N remote which is covered by the elro_800_switch module. Apparently I have the same issues there. The average pulse length was 300, but I had to change it to ~500 to make my remote working.

Am I the only one with such issues? Does it depend on the used receiver/low-pass filter?

CurlyMoo:
It depends on the input we get from users.

dominikkarall:
Ok, in that case I'm going to rewrite my code again, as I think that the binToDec function might also work if I change the average pulse length.
I'm going to send you another update...

wo-rasp:
May be that the values are different for the RCS 1000SN and for the RCS 1000N.
In this case, I propose to specify a separate protocol. Please post the Modell number of your switches
As discussed earlier, Brennstuhl switches themselves do not like the repetition factor of 10 for the protocol, based on my tests 4 is by far a much more suitable value !

dominikkarall:
Update to a single commit.
 - The binToDec function doesn't work for the RSL366 encoding, therefore I implemented this change
 - Minimum pulse length is to high in my case. 375 works great.
 - Average pulse length didn't work for me either. 480 works great.

@wo-rasp I will open another pull-request for RCS 1000SN (not N!).

wo-rasp:
I do own the RCS 1000 SN, and the values are okay for me. It is just the default value of 10 for txrpt that i would recommend to reduce to a value below 9, preferably 4.

CurlyMoo:
> I do own the RCS 1000 SN, and the values are okay for me. It is just the default value of 10 for txrpt that i would recommend to reduce to a value below 9, preferably 4.

@dominikkarall can you change this?

dominikkarall:
@CurlyMoo the RCS 1000 SN change is for another protocol. I would like to keep that change separate. This PR should only fix the RSL366T remote parsing.

dominikkarall:
@CurlyMoo is there anything else I need to do to get the pull request merged? It's only changing the receiving part, so sending part of the code remains completely untouched. Let me know if you need more input from my side. Thx.

PR #215, opened 2015-12-13
Add support for GT-9000 protocol

dominikkarall:
This pull request contains basic GT-9000 (GT-FSI-06 switch) support. GT-9000 protocol seems to be encrypted and currently I couldn't find any details about the encryption.

Protocol supports sending and receiving
 - systemcode (bits 0-3)
 - payload (bits 4-15)
 - ON/OFF combinations (bits 16-19)
 - unit (bits 20-23)
 - footer (bit 24)

State is not supported yet, as the state seems to be encoded in the payload part. GT-9000 remotes sends 4 different payloads for each button.
It's possible to use one of the 4 payloads to program the switch and use the other 3 to control the switch as well. It would also be possible to do button detection based on the 4 payloads in fhem or others.

I hope that this basic implementation helps others and maybe somebody has an idea to decode the payload as well.

### DIFF
--- a/libs/pilight/protocols/433.92/quigg_gt9000.c
+++ b/libs/pilight/protocols/433.92/quigg_gt9000.c
@@ -1,0 +1,394 @@
+/*
+	Copyright (C) 2015 CurlyMoo & dominikkarall
+
+	This file is part of pilight.
+
+	pilight is free software: you can redistribute it and/or modify it under the
+	terms of the GNU General Public License as published by the Free Software
+	Foundation, either version 3 of the License, or (at your option) any later
+	version.
+
+	pilight is distributed in the hope that it will be useful, but WITHOUT ANY
+	WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+	A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with pilight. If not, see	<http://www.gnu.org/licenses/>
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+
+#include "../../core/pilight.h"
+#include "../../core/common.h"
+#include "../../core/dso.h"
+#include "../../core/log.h"
+#include "../protocol.h"
+#include "../../core/binary.h"
+#include "../../core/gc.h"
+#include "quigg_gt9000.h"
+
+#define PULSE_QUIGG_SHORT	400
+#define PULSE_QUIGG_LONG	1100
+#define PULSE_QUIGG_FOOTER1	3000
+#define PULSE_QUIGG_FOOTER2	7000
+
+#define NORMAL_REPEATS		4
+#define AVG_PULSE_LENGTH	750
+#define RAW_LENGTH		50
+
+/* Encoding details:
+	Bit
+	0-3      First part of systemcode
+	4-19     Encrypted systemcode
+	16-19    ON/OFF statecodes (in encoded state, is also used decoded for systemcode)
+			ON2/OFF4        - 0000
+			OFF1/ON3/ONALL  - 1000
+			OFF2/ON4        - 0100
+			OFF2/ON4        - 1100
+			OFF1/ON3/ONALL  - 0010
+			ON1/OFF3/OFFALL - 1010
+			ON1/OFF3/OFFALL - 0110
+			ON2/OFF4        - 1110
+			ON1/OFF3/OFFALL - 0001
+			OFF2/ON4        - 1001
+			ON1/OFF3/OFFALL - 0101
+			OFF2/ON4        - 1101
+			ON2/OFF4        - 0011
+			OFF1/ON3/ONALL  - 1011
+			OFF1/ON3/ONALL  - 0111
+			ON2/OFF4        - 1111
+	20-24    Unit
+			Unit 1          - 1100
+			Unit 2          - 0101
+			Unit 3          - 1110
+			Unit 4          - 0111
+			ALL             - 0010
+	25       Footer (3000 7000)
+*/
+
+static int validate(void) {
+	if(quigg_gt9000->rawlen == RAW_LENGTH) {
+		if(quigg_gt9000->raw[quigg_gt9000->rawlen-1] >= (int)(PULSE_QUIGG_FOOTER2*0.9) &&
+		   quigg_gt9000->raw[quigg_gt9000->rawlen-1] <= (int)(PULSE_QUIGG_FOOTER2*1.1) &&
+		   quigg_gt9000->raw[quigg_gt9000->rawlen-2] >= (int)(PULSE_QUIGG_FOOTER1*0.9) &&
+		   quigg_gt9000->raw[quigg_gt9000->rawlen-2] <= (int)(PULSE_QUIGG_FOOTER1*1.1)) {
+			return 0;
+		}
+	}
+	return -1;
+}
+
+static void createMessage(int systemcode, int state, int unit) {
+	quigg_gt9000->message = json_mkobject();
+	json_append_member(quigg_gt9000->message, "id", json_mknumber(systemcode, 0));
+	if(unit==5) {
+		json_append_member(quigg_gt9000->message, "all", json_mknumber(1, 0));
+	} else {
+		json_append_member(quigg_gt9000->message, "unit", json_mknumber(unit, 0));
+	}
+	if(state == 1) {
+		json_append_member(quigg_gt9000->message, "state", json_mkstring("on"));
+	} else {
+		json_append_member(quigg_gt9000->message, "state", json_mkstring("off"));
+	}
+}
+
+static int decodePayload(int payload, int index) {
+	int ret = -1;
+	int hash[16] = { 0x0, 0x9, 0xF, 0x4, 0xA, 0xD, 0x5, 0xB,
+			 0x3, 0x2, 0x1, 0x7, 0xE, 0x6, 0xC, 0x8 };
+	
+	ret = payload^hash[index];
+	
+	return ret;
+}
+
+static int parseSystemcode(int *binary) {
+	int systemcode1dec = binToDecRev(binary, 0, 3);
+	int systemcode2enc = binToDecRev(binary, 4, 7);
+	int systemcode2dec = 0; //calculate all codes with base syscode2 = 0
+	int systemcode3enc = binToDecRev(binary, 8, 11);
+	int systemcode3dec = decodePayload(systemcode3enc, systemcode2enc);
+	int systemcode4enc = binToDecRev(binary, 12, 15);
+	int systemcode4dec = decodePayload(systemcode4enc, systemcode3enc);
+	int systemcode5enc = binToDecRev(binary, 16, 19);
+	int systemcode5dec = decodePayload(systemcode5enc, systemcode4enc);
+	int systemcode = (systemcode1dec<<16) + (systemcode2dec<<12) + (systemcode3dec<<8) + (systemcode4dec<<4) + systemcode5dec;
+
+	return systemcode;
+}
+
+static void pulseToBinary(int *binary) {
+	int x = 0;
+	for(x=0; x<quigg_gt9000->rawlen-1; x+=2) {
+		if(quigg_gt9000->raw[x+1] > AVG_PULSE_LENGTH) {
+			binary[x/2] = 0;
+		} else {
+			binary[x/2] = 1;
+		}
+	}
+}
+
+static void parseCode(void) {
+	int binary[RAW_LENGTH/2], state = -1;
+
+	pulseToBinary(binary);
+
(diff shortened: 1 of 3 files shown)


libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,241 @@
+/*
+	Copyright (C) 2015 dominikkarall
```
CurlyMoo: Can you keep my name here.

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,241 @@
...
+			 quigg_gt9000->raw[quigg_gt9000->rawlen-2] <= (int)(PULSE_QUIGG_FOOTER1*1.1)) { 
+		return 0;
+		}
+	}
+	return -1;
+}
+
+static void createMessage(int id, int payload, int unit, char *binary) {
+	quigg_gt9000->message = json_mkobject();
+	if(binary!=0)
```
CurlyMoo: Can you follow pilight coding styles here. Spaces in the if and using accolades.

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,241 @@
...
+		json_append_member(quigg_gt9000->message, "binary", json_mkstring(binary));
+	json_append_member(quigg_gt9000->message, "systemcode", json_mknumber(id, 0));
+	json_append_member(quigg_gt9000->message, "payload", json_mknumber(payload, 0));
+	json_append_member(quigg_gt9000->message, "unit", json_mknumber(unit, 0));
+}
+
+static void parseCode(void) {
+	int binary[RAW_LENGTH/2], x = 0;
+	char binaryCh[RAW_LENGTH/2];
+	strcpy(binaryCh,"");
```
CurlyMoo: Rather memset both arrays.

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,241 @@
...
+	length = decToBinRev(unit, binary);
+	for(i=0;i<=length;i++) {
+		x = i*2 + 20*2;
+		if(binary[i] == 1) {
+			createOne(x, x+1);
+		}
+	}
+}
+
+static void createState(int state) {
```
CurlyMoo: When will this be done?

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,241 @@
...
+	if(json_find_number(code, "unit", &itmp) == 0)
+		unit = (int)round(itmp);
+	if(json_find_number(code, "all", &itmp) == 0)
+		all = (int)round(itmp);
+	if(json_find_number(code, "off", &itmp) == 0)
+		state=0;
+	else if(json_find_number(code, "on", &itmp) == 0)
+		state=1;
+
+	if(systemcode==-1 || (unit==-1 && all==0)) {
```
CurlyMoo: Spaces

CurlyMoo:
What's the idea of the binary characters?

dominikkarall:
I will update the branch to reflect your comments.

Due to the fact that I can't decode the encrypted part of the message, I thought it might make sense to display the binary characters as well. Maybe somebody else has an idea how to decode it based on that information.

CurlyMoo:
I will wait for the merge until the binary info is decoded as well.

dominikkarall:
I updated the branch and included 2 hardcoded codes from my remotes. Those are working and everybody can use those to program their switches.
All GT-9000 remotes should be properly interpreted. So you should be able to detect which button was pressed on the remote. This wasn't possible in my first commit.

pilight-send usage
```
        [quigg_gt9000]
         -n --codenumber=codenumber     program your switches with predefined codes
         -u --unit=unit                 control the device unit with this code
         -t --on                        send an on signal to device
         -f --off                       send an off signal to device
         -a --all                       command to all devices with this systemcode
         -s --systemcode=systemcode     control one or multiple devices with this systemcode
         -l --payload=payload           payload in hex format
         -w --statecode=statecode       statecode for the corresponding payload
```
If you use internal codes, you just need to select codenumber (currently 1..2), unit (1..4) and on/off/all.
Example
```
pilight-send -n 1 -u 1 -t
```
If you want to use your own codes, you have to specify the received systemcode/payload/statecode combination.
```
pilight-send -u 1 -t -s 5 -l 0xF4F -w 4
```

If somebody else has a GT-9000 remote, I would be please if you could post your codes on the forum: https://forum.pilight.org/Thread-QUIGG-GT9000-Globaltronics-ALDI

@CurlyMoo Currently I have no idea how to decoded the 12 bits as I can't find a pattern. Therefore I would be pleased if you could include the code in the development branch. Maybe I can find some pattern if I have more payloads from other users. Thanks. Of course, comments are welcome.

CurlyMoo:
Do default payloads not just work?

dominikkarall:
What do you mean by default payloads? Each remote has their own payloads. If you have 2 sets of switches (8 units) you need 32 payloads. As I have 2 sets I hardcoded my payloads in the module.

CurlyMoo:
I mean, a master key payload. E.g. a payload that just works for all of them.

If payloads are attached to a remote, than we can consider them ID's right?

dominikkarall:
Unfortunately there is no master key payload. You always need the combination systemcode/payload/statecode which is unique to each remote.

I would like to stick to payload, as it is the encrypted ID. Each remote has only one ID but 16 payloads which are generated out of the ID.

CurlyMoo:
So people are unable to use the GUI because the payloads are rolling codes?

dominikkarall:
I'm not familiar with the pilight GUI, but I coded it that way, that the rolling code is used out of the 16 combinations and is used randomly. That's working. If I stick to only one code, after some time it might be that the switch doesn't react, therefore I used the random method which doesn't need to save a state in pilight.

dominikkarall:
So if you use the "internal codes" you don't need to specify any systemcode/payload/statecode.

CurlyMoo:
> So if you use the "internal codes" you don't need to specify any systemcode/payload/statecode.

And how reliable would that be?

dominikkarall:
The codes are the ones the remote sends, so it's fully reliable with the internal ones.

CurlyMoo:
And the internal codes can be received by pilight?

dominikkarall:
Yes, they can be received and are also shown correctly in the JSON message.

CurlyMoo:
And that would make `pilight-send -n 1 -u 1 -t` right?

dominikkarall:
Exactly.
n = 1-2
u = 1-4

dominikkarall:
@CurlyMoo have you tested my patch? Any comments?

CurlyMoo:
Still wondering if we shouldn't drop everything except the n and u.

dominikkarall:
I understand that the other parameters are a little bit complex. I wouldn't drop the other things completely, as I would like to achieve following:
 - People should be able to program their devices with different codes than their remotes. This allows the usage of the remote for other things.
 - Remote codes should be properly decoded so that systemcode, ON/OFF and unit can be interpreted.
 - If they still want to stick to the remote codes, they should be able to send those own codes to the switch as well.

Possible solution:
 - Use -n and -u for programming the switches with predefined codes.
 - Parse codes from remote control and show only
   - Systemcode (bit 0-3) - used to differentiate 2 remotes for the user
   - Payload (bit 0-19) - used to send the same code afterwards
   - Unit (1-4)
   - State (0/1)
 - Send the same code as the remote sends by using only -u 1-4, -l (systemcode, state and statecode encoded in -l already). This makes it easier for the user to send the same code again.
What do you think?

CurlyMoo:
> Send the same code as the remote sends by using only -u 1-4, -l

This doesn't rime with:

> Payload (bit 0-19) - used to send the same code afterwards

Right?

dominikkarall:
It's currently not implemented that way, but I would change it to reduce the number of parameters for the send command.

CurlyMoo:
Yes, if that would provide the same functionality but easier it's better.

dominikkarall:
I think I managed to decrypt the payload part :)
I need to do some tests with my switches to see if it works. So I hope to have an update this weekend.

dominikkarall:
Here we go!
Finally a working implementation of the GT-9000 protocol including decoding of the payload part.

Usage:
´´´
Use pilight-receive to decode your remote ID.
pilight-send -i <remote ID> -u <unit> -t
´´´
Some working IDs:
656022, 656023, 656071 and many others.

I think the Brennenstuhl RC 3600 is using a similar protocol as I found similar binary codes on the internet. So maybe somebody can get a proper hash table for RC 3600. I don't own that remote.

CurlyMoo:
So this is the final version?

dominikkarall:
It's the final version now.
My last commit was just a fix for \t in the pilight-send help text.

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,394 @@
...
+	else if(unit == 10)
+		unit = 2;
+	else if(unit == 7)
+		unit = 3;
+	else if(unit == 14)
+		unit = 4;
+	else if(unit == 4)
+		unit = 5;
+	else
+		return;
```
CurlyMoo: Can you add an error message here?

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,394 @@
...
+		else if(statecode == 10 || statecode == 6 || statecode == 1 || statecode == 5)
+			state = 0;
+		else
+			return;
+	} else {
+		return;
+	}
+
+	if(unit == 3 || unit == 4)
+		state = state?0:1;
```
CurlyMoo: Please add spaces here.

libs/pilight/protocols/433.92/quigg_gt9000.c
CurlyMoo: Nope, it's fine like this.

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,394 @@
...
+	if(json_find_number(code, "off", &itmp) == 0)
+		state=0;
+	else if(json_find_number(code, "on", &itmp) == 0)
+		state=1;
+
+	if((systemcode == -1) || (unit == -1 && all == 0)) {
+		logprintf(LOG_ERR, "quigg_gt9000: insufficient number of arguments");
+		return EXIT_FAILURE;
+	} else if(systemcode == -1 || systemcode < 650000) {
+		logprintf(LOG_ERR, "quigg_gt9000: invalid id range (recommended >650000)");
```
CurlyMoo: What do you mean by recommended? An systemcode is supported or not right?

libs/pilight/protocols/433.92/quigg_gt9000.c
```diff
@@ -0,0 +1,394 @@
...
+		} else if(unit == 2 || unit == 4) {
+			if((state == 1 && unit == 2) || (state == 0 && unit == 4))
+				statecode = on2[rand() % 4];
+			else
+				statecode = off2[rand() % 4];
+		} else if(unit == 5) {
+			if(state == 1)
+				statecode = off1[rand() % 4];
+			else
+				statecode = on1[rand() % 4];
```
CurlyMoo: Are you sure it works like this. What if the same number pops up each time? Shouldn't this be rolling?

libs/pilight/protocols/433.92/quigg_gt9000.c
dominikkarall: I'm going to remove the <650000 check, as I think that remotes could also have codes <650000. I have only 2 remotes which both have codes >650000.

libs/pilight/protocols/433.92/quigg_gt9000.c
dominikkarall: Is there a possibility to save the state to have a rolling code in plight? The same code is working as well, that's not a problem, but I want to imitate the remote as much as possible.

CurlyMoo: As long as pilight is running yes, create a linked list (check the weather station protocols for it) storing the values of a specific unit/id combination.

dominikkarall: I'll use a fix value, as the switch has absolutely no problem when using always the same code.

dominikkarall:
Let me know if the changes are ok, if so, I'm going to make a single commit.

CurlyMoo:
Seems fine

dominikkarall:
Ok. Updated to single commit.

CurlyMoo:
Helps to be stubborn to not merge half protocols :wink:

dominikkarall:
:stuck_out_tongue_winking_eye:  don't do that again, it was really stressful to find the bits during x-mas holidays :smile:

PR #224, opened 2016-01-20
Allow hyphens and underscores in sendmail TO option

NiekD:


CurlyMoo:
I don't think the removal of the `.` is ok. Gmail allows dots as alias. E.g. `curly.moo1` is the same as `curlymoo1`

NiekD:
I didn't know that. I will put it back in and create a new pull request.


Op 20-1-2016 om 18:59 schreef CurlyMoo:

NiekD:
I think the original . in the regex was wrong. A  single . on its own in a regex represents any character. To allow a . character only I will put in \\. instead. Do you agree?

CurlyMoo:
Just test it in a small unit test.

NiekD:
I did the test and it is working the way I suggested, but I realized that more special characters are officially allowed, bot in the local and in the domain part of the e-mail adress. Also nowadays domains like ".amsterdam" exist where the current regex only allows 2 or 3 characters.  Also if the regex rejects the e-mail addres, the error message just tells that the option "must contain an e-mail address" and not what is actually wrong.

Now we could create a rather complicated regex to cover all of this, but I don't think that this is really necessary.  Entering an illegal e-mail adrress will not make pilight crash; the message will simply not be delivered and this will be logged. 

To prevent that users with uncommon but (future) valid e-mail address will encounter problems (which then have to be fixed again) my idea now is to simplify the regex such that it only enforces the general format aaaaa@bbbbb.ccccc. Where aaaaa, bbbbb and ccccc can be any string of any length. 

This by the way also applies to the regex of the smtp-sender setting in the config.

What is your opinion ?

CurlyMoo:
> What is your opinion ?

Good idea :)

NiekD:
OK. I created and tested a regex that verifies that the e-mail address has the format:

aaaaa@bbbbb.ccccc

aaaaa, bbbbb and ccccc may have any length and may contain any character except @. Multiple adjacent dots are not allowed. (so aaaaa@bbbb.x.ccccc is valid, but aaaaa@bbbbb..ccccc is not). No other limitations.

I will create a PR for this fix. 
 
It is required for both sendmail.c (TO option) and settings.c (smtp-sender setting). Because it  is the same fix for both, I was thinking of creating just one PR. 
Or do you prefer separate PR's this case too?

CurlyMoo:
Just combine them.

Op 24-1-2016 om 18:58 schreef NiekD:

NiekD:
Do you mean I can close this PR and create a new one with one commit for sendmail.c and one commit for settings.c?

CurlyMoo:
No, add both in this PR.

NiekD:
@CurlyMoo 
I finally managed to tidy up my pull request. I modified the e-mail address validation in sendmail.c and settings.c, both in one commit.

yablacky:
Wait before pull in that:
The RE does not yet what claimed:
[^@^\\.\\.]
does not prevent two adjacent dots. It defines a set of all chars except @ the ^ and . (it does no harm that excluding the dot is twice in the exclude list).

yablacky:
Adjacent dots in an email's local part (left hand side from @) are unusual but not forbidden. The exact RE for the local part is very complicated so just deny @ does it good enough. To accept all public email domains the RE can be like this (blanks are for better reading - do not include them in the code - and don't forget to escape the backslash in the code):

^ [^@]+ @ ( [a-z\d] [a-z\-\d]* [a-z\d]) (\.[a-z\d] [a-z\-\d]* [a-z\d] )+$

NiekD:
@CurlyMoo 
I don't agree that the regex doesn't prevent two adjacent dots. I tetsted it once more to be sure. 

It was my intentention to make the regex more tolerant, so that we wouldn't have to change it every time some new/unusual address/domain name would be used. As said before a bad e-mail adress does not harm pilight; the message simply will not be delivered and that will be logged.

Since @yablacky claims that adjacent dots are allowed too, I propose to leave that check out and only prevent the @.

yablacky:
The RE
[^@^\\.\\.]+@[^@^\\.\\.]+\\.[^@^\\.\\.]+
has a match for
a..b@c.d

yablacky:
I just looked at rfc5322 again and it seems I was wrong and adjacent dots in the local part are also not allowed, sorry. If this is worth being checked by pilight (and not check many other email-addr rules), the RE should of course detect it.

yablacky:
What about implementing a new function (in core/common) that checks email-addresses and call that function where an email-addr needs to be checked. Like this:

int is_email_addr_valid(const char *addr, int check_domain_against_mx = 0); // return true if valid, false otherwise.

The benefit is that the check is the same at all places and there is only one place where the check itself is implemented (the RE, how complicate that might be or look, would only be written at one place). An additional optional check could be done: check the domain name of the email for being
a really existing email domain (which by default should no be checked because it may take time).

NiekD:
@yablacky 

You are quite right that my RE doesn't work the way I intended. This proves two things:

1. My test wasn't adequate
2. I'm not an expert with regular expressions :)

I think that implementing an email validation function is a good idea, but it is up to Curlymo to decide. 

Since my knowledge of regualr expressions is too basic, creating such a function would certainly be better in your hands.

CurlyMoo:
A general functions is ok.

yablacky:
ok, I can do that.

@NiekD
Should I also change the places where to call this new function? Or do you like it to change them yourself (it would rise your contribution count, if you care about that)?

NiekD:
@yablacky 
If you could do that, yes please. I don't care that much about my contribution count.

CurlyMoo:
What the status of that global function?

PR #226, opened 2016-01-25
Fix potential NULL pointer access in json_strdup().

yablacky:
If out-of-memory happend in json_strdup() the application crashed due to a
segmentation fault rather then terminate regularly. This has been fixed.


CurlyMoo:
Thanks

PR #227, opened 2016-01-25
Fix incomplete buffer initialization

yablacky:
Fix memset-length parameter when initializing buffers that are arrays of elements larger than byte (multiply the length by sizeof(buffer[0]) or just use sizeof(buffer)).

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -830,7 +830,7 @@ static int send_queue(struct JsonNode *json, enum origin_t origin) {
 				}
 				jprotocol = jprotocol->next;
 			}
-			memset(raw, 0, MAXPULSESTREAMLENGTH-1);
+			memset(raw, 0, sizeof(raw));
 			protocol->raw = raw;
 			if(match == 1 && protocol->createCode != NULL) {
 				/* Let the protocol create his code */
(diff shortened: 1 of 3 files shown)


CurlyMoo:
What's the difference?

yablacky:
The difference is like this:
int buffer[MAXPULSESTREAMLENGTH];
(A) memset(buffer, 0, MAXPULSESTREAMLENGTH);
versus
(B) memset(buffer, 0, sizeof(buffer));

The difference is that sizeof(buffer)  is not equal to MAXPULSESTREAMLENGTH.
Because sizeof(buffer) == sizeof(int)*MAXPULSESTREAMLENGTH.
With (A) only the first MAXPULSESTREAMLENGTH bytes(!) in the buffer are set to 0. Remaining buffer elements remain uninitialized.
With (B) the complete buffer with all int elements are initialized to 0.

CurlyMoo:
Aha, can you split the buffer issue with the NULL pointer issue.

yablacky:
ok, I will cancel this and make another PR "fix-too-short-buffer-initialization" in a few minutes.

PR #228, opened 2016-01-25
Fix pilight send crash on clarus protocol

yablacky:
Fix pilight-send crash on clarus_switch protocol if called with a very (e.g. 80 chars) long --id. In this case an on-stack buffer overflow happens. This has been fixed.


CurlyMoo:
Can you split the commits, so only clarus in this one. Can be done without having to reopening it.

yablacky:
ok, I cancel this and create another PR with only clarus crash issue.

PR #231, opened 2016-01-30
Fix timezone calculations on _WIN32 and on thread race conditions.

yablacky:
* On _WIN32 builds the result of localtime() and gmtime() calls were ignored.
  This has been fixed.
* Fixed calls to atomiclock(): some were too late (e.g. after getenv("TZ"))
  which was a bug, and some were unnecessary early (e.g. before initialization
  of local variables) which at least may block other threads a little bit
  longer then needed.
* Minor: Prevent succeeded calls of same fxn (getenv in this case).

### DIFF
--- a/libs/pilight/core/datetime.c
+++ b/libs/pilight/core/datetime.c
@@ -475,6 +475,23 @@ static pthread_mutex_t mutex_lock;
 static pthread_mutexattr_t mutex_attr;
 static unsigned short mutex_init = 0;
 
+#define out_of_memory() do {                    \
+                fprintf(stderr, "Out of memory.\n");    \
+                exit(EXIT_FAILURE);                     \
+        } while (0)
+
+
+static char *core_strdup(const char *str) {	// TODO: merge with json_strdup() etc. to a pilight_strdup()...
+	if(str == NULL) {
+		return NULL;
+	}
+	char *dup = malloc(strlen(str) + 1);
+	if(dup == NULL) {
+		out_of_memory();
+	}
+	return strcpy(dup, str);
+}
+
 void datetime_init(void) {
 	pthread_mutexattr_init(&mutex_attr);
 	pthread_mutexattr_settype(&mutex_attr, PTHREAD_MUTEX_RECURSIVE);
@@ -574,14 +591,14 @@ char *coord2tz(double longitude, double latitude) {
 time_t datetime2ts(int year, int month, int day, int hour, int minutes, int seconds, char *tz) {
 	logprintf(LOG_STACK, "%s(...)", __FUNCTION__);
 
-	atomiclock();
  	time_t t;
  	struct tm tm = {0};
-	char oritz[64];
-	memset(oritz, '\0', 64);
+	char *oritz = NULL;
 
-	if(getenv("TZ") != NULL) {
-		strcpy(oritz, getenv("TZ"));
+	atomiclock();
+
+	if((oritz = getenv("TZ")) != NULL) {
+		oritz = core_strdup(oritz);	// make copy because setenv() renders the pointer invalid.
 	}
 
 	tm.tm_sec = seconds;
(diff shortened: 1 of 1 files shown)


libs/pilight/core/datetime.c
```diff
@@ -577,11 +577,11 @@ time_t datetime2ts(int year, int month, int day, int hour, int minutes, int seco
 	atomiclock();
  	time_t t;
  	struct tm tm = {0};
-	char oritz[64];
+	char oritz[64], *s;
```
CurlyMoo: Can you initialize this pointer to NULL.

libs/pilight/core/datetime.c
```diff
@@ -612,22 +612,26 @@ time_t datetime2ts(int year, int month, int day, int hour, int minutes, int seco
...
+	if (strcmp(tz1, tz2) == 0) {	// prevent heavy calculations on simple case
+		return 0;
+	}
+
 	atomiclock();
 
 	time_t utc, tzsearch, now;
 	struct tm tm;
-	char oritz[64];
+	char oritz[64], s;
```
CurlyMoo: In the previous fix this was a pointer, why not now?

libs/pilight/core/datetime.c
```diff
@@ -680,12 +684,12 @@ int ctzoffset(void) {
 int isdst(time_t t, char *tz) {
 	logprintf(LOG_STACK, "%s(...)", __FUNCTION__);
 
-	char oritz[64];
+	char oritz[64], s;
```
CurlyMoo: Same here.

libs/pilight/core/datetime.c
```diff
@@ -612,22 +612,26 @@ time_t datetime2ts(int year, int month, int day, int hour, int minutes, int seco
...
-	if(getenv("TZ") != NULL) {
-		strcpy(oritz, getenv("TZ"));
+	if((s = getenv("TZ")) != NULL) {
+		strcpy(oritz, s);
 	}
 
 	now = time(NULL);
 #ifdef _WIN32
-	localtime(&now);
+	tm = *localtime(&now);
```
CurlyMoo: Is the output used anywhere?

libs/pilight/core/datetime.c
yablacky: it is wrong to save pointers of getenv() return values for alonger time (in particular if the environment will be modified later on, as is the case here) because the saved pointer became invalid after environment modifications. If you need the value later, you must store it in an own buffer (as oritz). There is still a bug if the TZ value in the environment is longer than be stored in oritz (64 chars).This could be fixed by allocating a buffer e.g. useing strdup.

It is ok to save the return value of getenv() to prevent an immediately followin getenv() call for the same variable. This is, what I did. Better is to set s=NULL after usage to prevent accidental use of s. I will chage that.

libs/pilight/core/datetime.c
yablacky: of course

libs/pilight/core/datetime.c
yablacky: I didn't check if the output is used. But in the #else case (e.g. _WIN32 not defined) the variable tm is being set with the localtime result whether or not tm is used later. If tm is not used, the code setting it up could be removed et all. You you want it this way?

libs/pilight/core/datetime.c
yablacky: ok, I see what you mean, there is a * missing....

libs/pilight/core/datetime.c
yablacky: There are even more declared but unused variables. I will clean that.

yablacky:
I squashed the commits into one.

CurlyMoo:
I need to dive into this one better.

CurlyMoo:
I manually merged these fixes in `rewrite`.

PR #232, opened 2016-01-31
Fix buffer overflow in setenv() and unsetenv() functions.

yablacky:
A buffer allocated by setenv() and unsetenv() was one byte too short.
If setenv() was called with option to not overwrite an existing variable
and that variable does not exist, then a NULL pointer was used and a
segfault happens (if the variable exists, it was unnecessary set again
to its current value). All this has been fixed.

Further the parameters passed to setenv() and unsetenv() are now checked
against NULL pointers and the functions do the best they can do without
crashing (return with failure indicator and errno set). This introduced
a feature for setenv() that, if called with a variable name but a value
of NULL, it does an unsetenv().

### DIFF
--- a/libs/pilight/core/common.c
+++ b/libs/pilight/core/common.c
@@ -161,10 +161,17 @@ int check_instances(const wchar_t *prog) {
 }
 
 int setenv(const char *name, const char *value, int overwrite) {
-	if(overwrite == 0) {
-		value = getenv(name);
+	if(name == NULL) {
+		errno = EINVAL;
+		return -1;
 	}
-	char c[strlen(name)+strlen(value)+1];
+	if(overwrite == 0 && getenv(name) != NULL) {
+		return 0; // name already defined and not allowed to overwrite. Treat as OK.
+	}
+	if(value == NULL) {
+		return unsetenv(name);
+	}
+	char c[strlen(name)+1+strlen(value)+1]; // one for "=" + one for term zero
 	strcat(c, name);
 	strcat(c, "=");
 	strcat(c, value);
@@ -172,7 +179,11 @@ int setenv(const char *name, const char *value, int overwrite) {
 }
 
 int unsetenv(const char *name) {
-	char c[strlen(name)+1];
+	if(name == NULL) {
+		errno = EINVAL;
+		return -1;
+	}
+	char c[strlen(name)+1+1]; // one for "=" + one for term zero
 	strcat(c, name);
 	strcat(c, "=");
 	return putenv(c);


CurlyMoo:
I'm really gratefull about all work you are doing in auditing the pilight code. Can we however agree on one thing and that is, don't close PR's but always fix them by using tools as rebase. I already hate it that pilight has so many opened and closed PR's done by people lot less skilled than you are.

That being said. Can we make sure that all pilight code is written the same way. Not:
```
if(...)
{
...
}
```
but:
```
if(...) {
...
}
```
Can you resolve this in this PR and then (without closing / reopening the PR) make everything a single commit again.

yablacky:
Yes, sorry. I chance position of the {}'s.
To be honest, I haven't no skills in using git so far (I know that you didn't want such people to contribute - but I couldn't resist). Sorry about deleting branches while leaning how to split commits etc. Hopefully doing it better now.

BTW: Do you have a preference for how to comment functions? I've not seen any in the code yet (I did not look into every module, though). My proposal would be to do it like in php-doc:
/**
 * Description of the_function.
 * @param type name description.
 * @return type description.
 */
int the_function(....) {
}

CurlyMoo:
I don't comment much on functions. IMHO functions should be self-explanatory. When needed i give some general advice.

Also, you can always test git on your own (test) repository. Also doing PR's can be easily done on your own test locations before doing PR's to pilight. That's the best way to learn.

So i will merge your fixes when you fixed the programming style.

yablacky:
I agree that functions should be self-explanatory; at least their name (in so far, e.g. isNumeric() is unfortunately poor named or is returning the "wrong" value, a 0 for ok which is false...).

Now sometimes meaning of function _parameter_ values are not (or can't be) self-explanatory. E.g. if pointers are passed, which is a common case in C, then a NULL may be forbidden or may have special meaning. I would like to document on that.

CurlyMoo:
Great. You have to learn something as well :+1: 

In C programming and in Linux in general the `0` is considered a True. That's also quite logical because you can normally only have one 'success' considition. Failure can have various reasons. So you can use the `< 0` numbers as hard failure and the `> 0` for success with additional information, such as a return length.

yablacky:
Thanks :-) I agree that the `0` (non-negative values in general) are used as a success indicator by for example lot of C standard libraries. I also agree for the reason about this interpetation of return values.

In contrast, the C language itself treats a 0 value as false and any other value as true.

Sometime it's hard to see which rule a self-written function follows. I found that code like this
```c
if(is_something_ok(something)) {
    // it is OK
} else {
    // it is not OK
}
```
is very readable an concise. Function that return a lib-style return value should be named differently, more like this:
```c
if(check_something(something) == 0) {
    // is OK 
} else {
    // not ok
}

// can be used like this:
if((err = check_something(something)) >= 0) {
    // is OK, with additional info in err.
} else {
    // not ok and the reason is err.
}
```
I think it would be a good idea to at least give new functions a name that has a hint of how to interpret its return value. How do you think?

Given this, my proposal in the email-addr-check issue in different pilight PR, that function should be
```c
int check_email_addr(const char* addr, ....etc... ); // returns >=0 on ok, <0 on err.
// err < 0 indicates what was wrong.
// err >= 0 indicates success (and for example if addr is even a list, the number of addresses -1). So if you don't accept lists, you may check for == 0. -- just an idea....
```

CurlyMoo:
That's also the reason why i keep my code as verbose as possible. I try to avoid the
```
if(something(...)) {
   ...
} else {
   ...
}
```
because it doesn't tell me anything about the variable types that the function returns.
`if(name == NULL)` and `if(name == 0)` are 99% the same thing, however, the `NULL` tells me we are comparing a pointer, and the `0` tells me we are comparing a numeric value.

yablacky:
ok, I accept that and follow your rules.

Finally, is it ok to you, when using the proposed php-style in case I really want to add comments to a function or function parameters?

CurlyMoo:
> Finally, is it ok to you, when using the proposed php-style in case I really want to add comments to a function or function parameters?

Ofc.

Can you still squash the commits into one single commit?

yablacky:
Done.

PR #233, opened 2016-02-01
Fix checking e-mail addresses; Added a common function to check them.

yablacky:
Previous check of e-email addresses was too strict.
This has been fixed, they are now checked more precise.

Implemented a common function check_email_addr() that can optionally
also accept lists of e-mail addresses. The option to check e-mail
domain names against DNS is prepared but ignored in the moment.

### DIFF
--- a/libs/pilight/core/common.c
+++ b/libs/pilight/core/common.c
@@ -49,6 +49,7 @@
 	#include <sys/mount.h>
 	#include <sys/ioctl.h>
 #endif
+
 #include <dirent.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -111,7 +112,7 @@ void array_free(char ***array, int len) {
 	}
 }
 
-unsigned int explode(char *str, const char *delimiter, char ***output) {
+unsigned int explode(const char *str, const char *delimiter, char ***output) {
 	if(str == NULL || output == NULL) {
 		return 0;
 	}
@@ -1018,4 +1019,94 @@ int file_get_contents(char *file, char **content) {
 	}
 	fclose(fp);
 	return 0;
+}
+
+char *str_trim_right(char *str, const char *trim_away) {
+	char *end = str;
+	if (str != NULL && trim_away != NULL && *trim_away != '\0') {
+		for (end += strlen(str); end > str && strchr(trim_away, end[-1]) != NULL; --end)
+			;
+		*end = '\0';
+	}
+	return str;
+}
+
+int checkdnsrr(const char *domain, const char *type)
+{
+	logprintf(LOG_STACK, "%s(%s,%s)", __FUNCTION__, domain ? domain : "(NULL)", type ? type : "(NULL)");
+	return 0;	// NYI.
+}
+
+/**
+ * Check if an email address is valid.
+ * @param char* address The email address to check. Leading+trailing white spaces are accepted.
+ * @param bool allow_lists If true, addr is allowed to be a comma separated list of email addresses.
+ * @param bool check_domain_can_mail If true, check if domain exists in DNS system and accepts email.
+ * @return int Values < 0 indicate error. Values >= 0 indicate success.
+ * Where -1: syntax error, -2: DNS check for a domain failed, -8: severe internal error.
+ */
+int check_email_addr(const char *address, int allow_lists, int check_domain_can_mail) {
+	if(address == NULL) {
+		return -1;
+	}
+
+	#define ALPHANUMERICS "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
+
+	static const char white_spaces[] = " \t";
+	static const char local_valid_chars[] = ALPHANUMERICS ".!#$%&'*+-\\/=?^_`{|}~";
+	static const char domain_valid_chars[] = ALPHANUMERICS ".-"; // . and - are not allowed everywhere.
+
+	const char *at = NULL, *domain_name = NULL;
+	char *addr = NULL;
+	char **addr_list = NULL;
+	size_t addr_count = explode(address, ",", &addr_list), ii = 0;
+	int ret = -1;
+	for(ii = 0; ii < addr_count; ii++) {
+		ret = -1;	// assume error
+		if (allow_lists == 0 && addr_count > 1) {
+			break;	// only one address allowed.
+		}
+		addr = addr_list[ii];
+		addr += strspn(addr, white_spaces);	// trim left
+		str_trim_right(addr, white_spaces);
+
+		// Now check addr. First some quick pre-checks:
+		at = strchr(addr, '@');
+		if(at == NULL || strstr(addr, "..") != NULL) {
+			break;	// no @ found or two consecutive dots found somewhere.
+		}
+		domain_name = at+1;
+
+		if(at == addr || at[-1] == '.' || addr[0] == '.') {
+			break;	// local part emtpy, or starts or ends with a dot.
+		}
+		if(addr+strspn(addr, local_valid_chars) < at) {
+			break;	// local part contains invalid chars.
+		}
+		if(strchr(domain_name, '.') == NULL || strlen(domain_name) < 5) {
+			break;	// not at least one dot in domain or domain is not at least 2 x 2 chars.
+		}
+		if(strchr(".-", domain_name[0]) != NULL || strchr(".-", domain_name[strlen(domain_name)-1]) != NULL) {
+			break;	// domain starts or ends with a dot or dash.
+		}
+		if(strstr(domain_name, ".-") != NULL || strstr(domain_name, "-.") != NULL) {
+			break;	// a dot delimited domain part starts or ends with a dash.
+		}
+		if(domain_name[strspn(domain_name, domain_valid_chars)] != '\0') {
+			break;	// domain contains invalid chars.
+		}
+
+		// Finally check domain name against DNS if desired:
+
+		if(check_domain_can_mail != 0 && checkdnsrr(domain_name, "MX") < 0) {
+			ret = -2;
+			break;	// domain not found or does not accept emails.
+		}
+
+		ret = 0; // Address ok (and all ok if this is the last one in the list).
+	}
+	array_free(&addr_list, addr_count);
+	return ret;
+
+	#undef ALPHANUMERICS
 }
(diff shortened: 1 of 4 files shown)


libs/pilight/core/common.c
```diff
@@ -49,6 +49,11 @@
 	#include <sys/mount.h>
 	#include <sys/ioctl.h>
 #endif
+
+#ifndef _WIN32
```
CurlyMoo: Regex also doesn't work on FreeBSD.

libs/pilight/core/common.c
```diff
@@ -1019,3 +1024,161 @@ int file_get_contents(char *file, char **content) {
...
+}
+
+#if defined(__FreeBSD__) || defined(_WIN32)
+#define use_regexp 0
+#else
+#define use_regexp 1
+#endif
+
+// Do NOT write the next as range A-Za-z0-9 because the string may be used without RE.
+#define AtoZ_atoz_0to9 "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
```
CurlyMoo: Why not name this `ALPHANUM`.

Also it's a good practice to write defines in capitals.

libs/pilight/core/common.c
```diff
@@ -1019,3 +1024,161 @@ int file_get_contents(char *file, char **content) {
...
+int check_email_addr(const char *addr, int allow_lists, int check_domain_can_mail) {
+	if(addr == NULL) {
+		return -1;
+	}
+#if use_regexp
+	const regex_t *RE = NULL;
+	int reti = 0;
+
+	static const char email_pattern[] =
+		"^[" email_local_part_valid_chars "]+"
```
CurlyMoo: Why don't you integrate the `email_local_part_valid_chars` directly here?

libs/pilight/core/common.c
```diff
@@ -1019,3 +1024,161 @@ int file_get_contents(char *file, char **content) {
...
+		")$";
+
+	if((RE = check_email_compiled_RE) == NULL) { // first check without lock.
+		// RE not compiled yet.
+		atomiclock();
+		if((RE = check_email_compiled_RE) == NULL) { // locked now, need to check again.
+			reti = regcomp(&check_email_common_RE, email_pattern, REG_EXTENDED);
+			if(reti == REG_NOERROR) {
+				check_email_compiled_RE = RE = &check_email_common_RE;
+				atexit(check_email_free_RE);
```
CurlyMoo: Let the main_gc call a garbage collection function in here.

libs/pilight/core/common.c
```diff
@@ -1019,3 +1024,161 @@ int file_get_contents(char *file, char **content) {
...
+				logprintf(LOG_ERR, "could not compile regex pattern '%s' to check email addresses: %s",
+						email_pattern, errmsg);
+				return -1;
+			}
+		}
+		atomicunlock();
+	}
+#else
+	static const char local_valid_chars[] = email_local_part_valid_chars;
+	static const char domain_valid_chars[] = AtoZ_atoz_0to9 ".-"; // . and - are not allowed everywhere.
```
CurlyMoo: Why not but the alphanummeric string directly here? Same counts for the `email_local_part_valid_chars`? Or when needed at the beginning of this function instead of a define?

libs/pilight/core/common.c
```diff
@@ -1019,3 +1024,161 @@ int file_get_contents(char *file, char **content) {
...
+#endif
+
+	char _addr[strlen(addr)+1];
+	const char *at = NULL, *domain_name = NULL;
+	char *comma = NULL, *next_addr = NULL;
+
+	next_addr = strcpy(_addr, addr);	// get a writable copy.
+	while(next_addr != NULL) {
```
CurlyMoo: You can also use the pilight `explode` function.

libs/pilight/core/common.c
```diff
@@ -1019,3 +1024,161 @@ int file_get_contents(char *file, char **content) {
...
+			return -1;	// no error but no match.
+		}
+		if(reti != REG_NOERROR) {
+			char errmsg[regerror(reti, RE, NULL, 0)];
+			regerror(reti, RE, errmsg, sizeof(errmsg));
+			logprintf(LOG_ERR, "could not execute regex to check email addresses: %s", errmsg);
+			return -8;
+		}
+#else
+		// Do some checks without RE engine (catch many but not all wrong syntax):
```
CurlyMoo: Isn't it possible to use a totally regex free system that works on both windows, linux, and BSD? You are implementing it already so better not do it redundant.

libs/pilight/core/common.c
yablacky: I'm wondering because there are lots of places like
```´c
#ifndef _WIN32
#include <regex.h>
```´
FreeBSD is accounted for later in the file, near ```#define use_regex```

File: libs/pilight/core/common.c (hunk 2 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: yes, but in this case capitals would lead to ATOZ_ATOZ_... which ruins the information the name transports. I indeed already thought about ALPHANUM or similar. Finally I wanted it to look more similar to the RE range-expression that would be written inside ```[ ]```

Comment (CurlyMoo) [inline libs/pilight/core/common.c]: >  Finally I wanted it to look more similar to the RE range-expression that would be written inside [ ]

That's what we need to avoid. Having defines look similar as the rest.

File: libs/pilight/core/common.c (hunk 3 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: because email_local_part_valid_chars is used more than once in the code (in the use_regex part and the not use_regex part). When integrating the expression directly, there would be two places where this set of chars is written. If you change one it tends to forget the other. In particular because it is not even clear that it should be the same set of chars. Using a name for the set and define it at one place makes all this clear.

Comment (CurlyMoo) [inline libs/pilight/core/common.c]: But why not placing this information in the function itself instead of in a define. The whole point of the function in the first place is not to validate email addresses anywhere else right?

File: libs/pilight/core/common.c (hunk 4 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: ok

File: libs/pilight/core/common.c (hunk 5 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: yes, local_valid_chars could be eleminated, its used only once. I indeed already thought about not to define this buffer. Finally I introduced the name because it does not cost anything and has the benefit that there is a name for the charsets that can be examined when debugging.

File: libs/pilight/core/common.c (hunk 6 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: yes I've seen that. But ```explode()``` allocates lots of extra memory buffers on the heap. I would have to care for out-out-mem and must free the stuff. Far to heavy for what is needed here. I considered to use strtok()... But since delimiter is always the comma it does not make things simpler.

File: libs/pilight/core/common.c (hunk 7 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: As I wrote it, I had the same idea :-) For the purpose needed in pilight, the regex-free checks could be good enough (and much faster, I guess). Should I eleminate the regex stuff? Would also make cleanup easier.

File: libs/pilight/core/common.c (hunk 3 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: Not shure what you mean.

The defines are not outsite of the check-fxn to ease validation of email addresses anywhere else (would this be my intention, I defined it in common.h). It's just because a ```#define ``` is not bound to function scope. Its not a big difference to write a #define in the function body or before. If you meant this, I can put the #defines inside. It just wouldn't change that the #defines remain for the rest of the file (unless #undef'd).

If you meant to have a (static and/or const) variable in the function body that holds the string snippets, the answer is: I didn't do that here because I make use of letting the compiler concatenate string literals (at compile time) out of the string snippets that are ```#defined```. Strings in variables must be concatenated at run time and would cause me to write code that does it (and it would be done on each call despite its always the same "constant" result).

File: libs/pilight/core/common.c (hunk 7 above)
Comment (CurlyMoo) [inline libs/pilight/core/common.c]: I know, but consider this is only done once at startup and what are we talking about. A few bytes and 0.0001sec?

File: libs/pilight/core/common.c (hunk 6 above)
Comment (CurlyMoo) [inline libs/pilight/core/common.c]: I know, but consider this is only done once at startup and what are we talking about. A few bytes and 0.0001sec?

File: libs/pilight/core/common.c (hunk 3 above)
Comment (CurlyMoo) [inline libs/pilight/core/common.c]: I mean:
```
void tmp(void) {
   char *alphanum = 'ABCDE.....';
   ...
}
```

**Comment by yablacky** [comment]
As proposed, I eleminated regex to check emails. I also applied other proposals (eleminating #defines and changed names, and there is no need for cleanup code).

File: libs/pilight/core/common.c (hunk 6 above)
Comment (yablacky) [inline libs/pilight/core/common.c]: I'm not convinced that explode() has benefits here; anyhow I changed the code to use explode().

File: libs/pilight/core/common.c [hunk 8]
```diff
@@ -111,7 +112,7 @@ void array_free(char ***array, int len) {
 	}
 }
 
-unsigned int explode(char *str, const char *delimiter, char ***output) {
+unsigned int explode(const char *str, const char *delimiter, char ***output) {
```
CurlyMoo: I don't believe explode is limited to `const char`. You can bypass error by doing:
```
const char *a = 'foo';
char *p = a;
explode(a, ...);
```

yablacky: Yes but this is not what ```const``` mean in function declarations. In function declarations ```const``` is a promise of that function that it will not modify the item declared as const (in this case the callers buffer of chars that str is pointing to). The compiler checks that promise and complains if explode tryies something like *str = 'x'.

The promise of the function not to modify items in a callers buffer has no impact of what the caller itself does with its buffer and the chars in it.

yablacky: BTW, your code is a good example of whats happen if code tries to break the 'const' promise:
```c
const char *a = "foo";  // its promised that further code won't modify chars that a points to.
char *p = a;    // compiler warns that 'const' attribute is dropped (this breaks the promise because p could be used to modify chars that a points to).
```

yablacky: Just to make that clear: changing a function from
```int f(char *p) { ... }```
to
```int f(const char *p) { ... }```
does not add an resctriction of how f() can be called. It adds an enhacement:
```c
int f1(char *p) { ... }
int f2(const char *p) { ... }

char a[] = "...";
const char b[] = "...";

f1(a);
f1(b);  // compiler warning! f1 is resticted to be called with non-const char buffer.
f2(a);
f2(b);  // no striction, no compiler warning.
```
On the other hand, changing a function from
```int *g() { ... }```
to
```const int *g() { ... }```
adds an restriction to how the return value can be stored:
```c
int *g1() { ... }
const int *g2() { ... }

int *a;
const int *b;

a = g1();
b = g1();
a = g2();  // compiler warning: return value is restricted to be stored in a const int* variable.
b = g2();
```

PR #238, opened 2016-02-13
fix buffer overflow in elro_300_switch protocol

cxxcoder:
Because the rawlen array has twice the size of the binary array there is currently a buffer overflow which can make pilight crash. This happened for my but the event was triggered by RF noise. Unfortunately I have no device using the elro_300_switch protocol so that I could test the changed code. But with this code there are no longer crashes.


CurlyMoo:
This was the previous implementation. Maybe that helps fixing:
https://github.com/pilight/pilight/blob/ac6d35279d03f46af02f378b9209e06f1aa34510/libs/protocols/elro_300.c#L66-L78

CurlyMoo:
Can you make this one commit?

cxxcoder:
I have merged it to one commit.

PR #242, opened 2016-02-20
fix alecto_wx500 protocol

cxxcoder:
There is a bug in the alecto_wx500 protocol I found by an analysis by valgrind.
RAWLENGTH is 74 and thus the binary array has a length of 37 but because of "rawlen-1" in the for loop bit 36 of the binary array is not set.
Because of this valgrind gives an "Conditional jump..." error in line 81 when this bit is used.
With this fix there were no valgrind errors for this protocol any more.


wo-rasp:
I agree, due to the < comparison, the last value of x is 71 (35th execution of the loop) and not 73.

PR #243, opened 2016-02-20
fix read beyond raw array in rsl366 protocol

cxxcoder:
There is a bug in the rsl366 protocol I found by an analysis by valgrind.
RAWLENGTH is 50 and thus rawlen-1 should be 49.
Thereby the for loop increments x until 48.
But then raw[x+3] will read beyond the raw array.
However, valgrind did not log invalid writes in line 70 or 72 which I cannot explain.
Perhaps someone with rsl366 hardware can test this fix.
I have no rsl366 hardware and, when I made the test, the event seemed to be triggered by RF noise.
With this fix there were no valgrind errors any more for this protocol.

### DIFF
--- a/libs/pilight/protocols/433.92/rsl366.c
+++ b/libs/pilight/protocols/433.92/rsl366.c
@@ -65,8 +65,8 @@ static void parseCode(void) {
 	int x = 0, i = 0, binary[RAW_LENGTH/4];
 
 	/* Convert the one's and zero's into binary */
-	for(x=0;x<rsl366->rawlen-1;x+=4) {
-		if(rsl366->raw[x+3] > (int)((double)AVG_PULSE_LENGTH*((double)PULSE_MULTIPLIER/2))) {
+	for(x=3;x<rsl366->rawlen;x+=4) {
+		if(rsl366->raw[x] > (int)((double)AVG_PULSE_LENGTH*((double)PULSE_MULTIPLIER/2))) {
 			binary[i++]=1;
 		} else {
 			binary[i++]=0;


wo-rasp:
I need to look it up when i am back home, but I think the rsl366->raw[] array is larger then RAWLENGTH, and that is the reason why valgrind is not reporting a violation.
In my opinion the code change in itself is correct and valid.
But in my opinion the real bug is that rsl366->rawlen may be used before it is initialized, as that happens in createCode only, however as the variable is used in validate, and validate can be called before createCode is called for the first time. Until that takes place, the value is undefined, therefore "rsl366->rawlen = RAW_LENGTH" must also be part of rsl366Init(), preferably in L211.

yablacky:
Don't get confused with handler callbacks used for sending and those used for receiving. When receiving, the rawlen value is setup by the "interrupt handler" that received the raw signal pulses. It does this before knowing which protocol handler is responsible. For each such set of received raw pulses, the daemon iterates over its known protocols and calls their validate() before possibly calling their parseCode(). So in parseCode() of a protocol handler, the rawlen is what its validate() allows to be (exactly RAW_LENGTH in this case).

Even if the old '''raw[x+3]''' was not wrong (takes brain contortions to validate it), I like your change because it makes things clear.

wo-rasp:
IMHO the original code is faulty, and the corrected code is okay.

I am less concerned with the raw array, my concern is the binary array, as that one is written to, outside of its boundary.

 

RAW Pulses are grouped in sets of 12 times 4 pulses plus 2 footer pulses.

A binary 0 is decoded whenever there is a low (e.q. short) 1st and low (short) 4th pulse in each group.

A binary 1 is decoded whenever there is a high (long) 2nd and high (long) 4th pulse in each group.

 

As it seems that the 1st pulse of a group is always low and the 2nd pulse is always high, that decoding step was omitted (we could argue about that one, but ….)

These 12 groups form the 12 binary bits, thus the last pulse to be decoded is raw[47].

 

The original code is executed with the last values x=48 and i=13.

The revised code is executed with its last value x=47 and i=12.

 

Von: yablacky [mailto:notifications@github.com] 
Gesendet: Sonntag, 21. Februar 2016 18:23
An: pilight/pilight
Cc: wo-rasp
Betreff: Re: [pilight] fix read beyond raw array in rsl366 protocol (#243)

 

Don't get confused with handler callbacks used for sending and those used for receiving. When receiving, the rawlen value is setup by the "interrupt handler" that received the raw signal pulses. It does this before knowing which protocol handler is responsible. For each such set of received raw pulses, the daemon iterates over its known protocols and calls their validate() before possibly calling their parseCode(). So in parseCode() of a protocol handler, the rawlen is what its validate() allows to be (exactly RAW_LENGTH in this case).

Even if the old '''raw[x+3]''' was not wrong (takes brain contortions to validate it), I like your change because it makes things clear.

PR #244, opened 2016-02-21
Fix raspberrypi initialization issues.

yablacky:


CurlyMoo:
If you have wiringX fixes, please do them in wiringX itself:
https://github.com/wiringx

PR #246, opened 2016-02-28
Add config-option to receive even self-sent messages.

yablacky:
The config option is at { "settings" : { "receive-while-sending": 0 or 1 } }
and the pilight-daemon is using it.  The purpose of this option is to test
protocol handler implementations like this:
Run pilight-receive at term one to see daemon logs. Then running a

  pilight-send -p proto --id xx --unit yy etc...

at term two should produce a sent and a receive message stating the same
id, unit and command codes. If they differ constantly the implementation
should be reviewed.

Note a strange thing: due to thread priorities, the receiving-log-message
may sometime appear before the send-log-message.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -164,6 +164,11 @@ static int bcqueue_number = 0;
 
 static struct protocol_t *procProtocol;
 
+/* By default the reveiver is turned to wait-mode while
+	sending to prevent receivin the sent data.
+	For debug reasons this might be helpful.  */
+static int receive_while_sending = 0;
+
 /* The pid_file and pid of this daemon */
 #ifndef _WIN32
 static char *pid_file;
@@ -671,6 +676,9 @@ void *send_code(void *param) {
 						json_append_member(message, "uuid", json_mkstring(sendqueue->uuid));
 					}
 					json_append_member(message, "repeat", json_mknumber(1, 0));
+					if(receive_while_sending != 0) {
+						json_append_member(message, "receive-while-sending", json_mknumber(receive_while_sending, 0));
+					}
 				}
 			}
 			if(sendqueue->settings != NULL && strcmp(sendqueue->settings, "{}") != 0) {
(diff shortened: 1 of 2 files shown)


woutput:
I like this too.
It may also help @Rschnauzer in [this thread] (https://forum.pilight.org/Thread-Verify-pilight-send-receive-function)

CurlyMoo:
Why shouldn't we enable this always?

Rschnauzer:
I am searching for a method to verify function of transmitter and receiver.
Several threads in the forum discuss about the compatibility and quality of transmitter and receiver.
I exchanged my transmitter and receiver but I still receive a lot of noise and telegrams of known sources with incomplete pulses.
A feature to verify the function of transmitter and receiver over the air would help to debug the installation.
Regards Rschnauzer

wo-rasp:
Most noise is generated due to the AGC (Automatic Gain Control) of the reciever. Check the drop in noise when you press the button on a remote control. Incomplete pulses could be either an indication that multiple RF sources transmit at the same time, or a bad adaption of the antenna (weak RF signal).

woutput:
@CurlyMoo said "Why shouldn't we enable this always?"

I don't see a reason to disable it, but I also don't know the current implementation in which this functionality apparently is (I think actively; by someone?) disabled (for some reason?)

CurlyMoo:
> I don't see a reason to disable it, but I also don't know the current implementation in which this functionality apparently is (I think actively; by someone?) disabled (for some reason?)

It was disabled by me ofc, which the idea that sending and receiving over the same antenna could be hurtfull (without any real further argumentation).

Rschnauzer:
With my installation the transceiver and receiver modul has separate antenna. Is it possible to connect the outlet of both modules to one antenna?
Regards Rschnauzer

CurlyMoo:
This feature has become obsolete in the latest code.

PR #248, opened 2016-03-01
Fix header footer issue for Dostman 30.3200 device.

wo-rasp:
Change is compatible with standard tfa and conrad weather device
https://forum.pilight.org/Thread-Protocols-with-Header-and-Footer?pid=17260#pid17260


CurlyMoo:
Please do the PR on either development or rewrite, not both. I can merge development into rewrite myself. This to make sure we don't have parralel mixed up branches.

PR #255, opened 2016-03-14
implement media access control by csma and csma/cd

andiwand:


andiwand:
this should maybe use your implementation of threadpools instead of creating its own thread.

CurlyMoo:
Will check it later. Currently working on wiringX.

PR #259, opened 2016-03-24
Support for SOENS Wheatherstation added to tfa protocol

wo-rasp:
https://forum.pilight.org/Thread-Soens-Innovation-TFA-similar-protocol?pid=17605#pid17605

Conflicts:
	libs/pilight/protocols/433.92/tfa.c
Conflict was with the revision number only.
I propose to keep the minor revision number in sync with the development branch.
1.2 on development --> 2.2 on the rwrite branch


CurlyMoo:
I will merge things from development to rewrite myself, That prevents commit conflicts. So if you PR'ed against development, don't do the same in rewrite, or only PR in rewrite.

PR #260, opened 2016-03-25
Add pointer check for prot->rawlen parameter to avoid buffer overflow…

wo-rasp:
… parsecode function

https://forum.pilight.org/Thread-Soens-Innovation-TFA-similar-protocol?pid=17613#pid17613


CurlyMoo:
You forgot to update the version numbers.

wo-rasp:
I have updated the version numbers for all 46 files affected.

PR #261, opened 2016-03-28
Updated s*mfy protocol stack

wo-rasp:
Some minor bugfixes were required in daemon.c, somfy_rts.c, protocol.h and protocol.c.
Tested with the config in the wiki and in an adhoc pilight environment and found operational from the CLI interface.
For the WEBGUI, the third button (MY) is an absolutely must have as that button stops the shutter movement. pilight.js in the somfy_release branch has a preliminary implementation of that button.
With 3 buttons you can master all daily operations.
For fancy interaction 5 buttons are required to implement UP+MY and DOWN+MY.


CurlyMoo:
I know about the GUI requirement. Can we make one S*mfy specific for now? So we implement a device type called S*mfy with it's own GUI?

Did the code work with my small changes?

wo-rasp:
Sure could we make a Smfy specific one, however the third button would not interfere with the other protocols using the screen interface, as all other protocols have no MY string they jump onto and react, so the change is, as far as i can judge it, compatible with all existing implementations.
The current version in somfy_release is compatible with all langauges define, including french.

The changes were not working out of the box, from daemon.c the pointer for the wakeup pulses needs to be passed by address (or reference), so that the adress to the external function's wakeup array is passed back to daemon.c, and the prototype definitions were missing.

CurlyMoo:
> Sure could we make a Smfy specific one, however the third button would not interfere with the other protocols using the screen interface, as all other protocols have no MY string they jump onto and react, so the change is, as far as i can judge it, compatible with all existing implementations.
The current version in somfy_release is compatible with all languages defined, including french.

Did your GUI implementation deal with this?

> The changes were not working out of the box, from daemon.c the pointer for the wakeup pulses needs to be passed by address (or reference), so that the adress to the external function's wakeup array is passed back to daemon.c, and the prototype definitions were missing.

Ofc, but with my changes which again were fixed by you :smile:

wo-rasp:
GUI:
Yes, that's how i dealt with it.
In function createScreenElement, the 8th line has the additional check:
sDevId+'_screen_stopped" value="my"
translating the my value into the "screen_stopped" checkboxradio interaction.

Independent of the revised implementation there is a general issue with the screen button GUI element:
whenever you press the buttons in a too fast sequence, one of the GUI button starts flickering and won't stop again.

I did not modify any logic, I addressed code and compiler related warnings only and tested the overall environment (CLI and 2 button GUI).

CurlyMoo:
Can you add the GUI in the PR again? I didn't notice the s_mfy specific checks.

wo-rasp:
Done. I had to increase the size in pilight.css to make it fit the text of english/french language.
THe workaround for the flashing button element is pressing F5 in Firefox.
I set the Version number to 1.0 and removed the comments of the development history.

libs/webgui/pilight.js
```diff
@@ -361,10 +361,11 @@ function createScreenElement(sTabId, sDevId, aValues) {
...
+			oTab.append($('<li  id="'+sDevId+'" class="screen" data-icon="false"><div class="name">'+aValues['name']+'</div><div id="'+sDevId+'_screen" class="screen" data-role="fieldcontain" data-type="horizontal"><fieldset data-role="controlgroup" class="controlgroup" data-type="horizontal" data-mini="true"><input type="radio" name="'+sDevId+'_screen" id="'+sDevId+'_screen_down" value="down" /><label for="'+sDevId+'_screen_down">'+language.down+'</label><input type="radio" name="'+sDevId+'_screen" id="'+sDevId+'_screen_stopped" value="my" /><label for="'+sDevId+'_screen_stopped">'+language.stopped+'</label><input type="radio" name="'+sDevId+'_screen" id="'+sDevId+'_screen_up" value="up" /><label for="'+sDevId+'_screen_up">'+language.up+'</label></fieldset></div></li>'));
```
CurlyMoo: Where do you hide the `my` buttons based on the protocol specifications. As in, don't show the buttons in regular screens?

wo-rasp:
Okay, integrated.
In principle there is an alignment issue with the buttons on the right side.
I assume it is caused by the framework and there is not an easy fix available,
a) It would be nice to have the buttons at the right side aligned to the right
b) When you display the interface on mobile devices, the buttons will overlap with the text on the left.

PR #267, opened 2016-04-17
Added unitec switch protocol

yablacky:


libs/pilight/protocols/433.92/unitec.c
```diff
@@ -0,0 +1,245 @@
...
+	unitec_switch->hwtype = RF433;
+	unitec_switch->minrawlen = RAW_LENGTH;
+	unitec_switch->maxrawlen = RAW_LENGTH;
+	unitec_switch->maxgaplen = MAX_PULSE_DIV;
+	unitec_switch->mingaplen = MIN_PULSE_DIV;
+
+	options_add(&unitec_switch->options, 't', "on", OPTION_NO_VALUE, DEVICES_STATE, JSON_STRING, NULL, NULL);
+	options_add(&unitec_switch->options, 'f', "off", OPTION_NO_VALUE, DEVICES_STATE, JSON_STRING, NULL, NULL);
+	options_add(&unitec_switch->options, 'u', "unit", OPTION_HAS_VALUE, DEVICES_ID, JSON_NUMBER, NULL, "^[0-5]$");
+	options_add(&unitec_switch->options, 'i', "id", OPTION_HAS_VALUE, DEVICES_ID, JSON_STRING, NULL, "^X[0-9A-Fa-f]+$");
```
CurlyMoo: What's with the `X`?

CurlyMoo:
Can you follow the pilight coding style as much as possible?

libs/pilight/protocols/433.92/unitec.c
yablacky: Other protocols name their --id A00 .. F63 or so. I don't know if unitec has a device where the id can be set via dip-switch or so and what would appoint the id's name and range. Mine has no such. To be as much as flexible I named it X with all the remaining bits.

yablacky:
i found one single ```if ()``` (with blank) and one single line indended by blanks not tabs. Do you mean this?

libs/pilight/protocols/433.92/unitec.c
CurlyMoo: But why not just use numbers?

libs/pilight/protocols/433.92/unitec.c
```diff
@@ -0,0 +1,245 @@
...
+	while(--c >= 0) {
+		unitec_switch->raw[    2*s  ]=MIN_PULSE_LENGTH;
+		unitec_switch->raw[1 + 2*s++]=MAX_PULSE_LENGTH;
+	}
+}
+
+static void createHigh(int s, int c) {
+
+	while(--c >= 0) {
+		unitec_switch->raw[    2*s  ]=MAX_PULSE_LENGTH;
```
CurlyMoo: In regard to coding style i also meant this.

libs/pilight/protocols/433.92/unitec.c
yablacky: Why not X?

CurlyMoo: Because it makes no sense as for as i understand. It doesn't add any value. Just using nummeric ID's would suit.

yablacky: ok, then I explain, why not just a number.

First, ist is not a numeric. The id is a string build from 16 different chars. I could made it a string out of any other number of chars, 8, 19 or 24 or 33 different chars for example. But I choose 16 because this leads to short enough IDs of max. 5 chars.
For example, with the X it could be ```XA1``` (I need that below).

Second, its very future proof. In case later someone found out and tells us that there are devices that have a particular named ID and that consist of - for example - a letter and digits, that can be added to the protocol without conflicts (unless these IDs also start with X, of course, but its foreseeable that that won't happen).

Lets assume we found out that unitec has devices that can be set to ID like "A1" by dib-switches. Then I would like to send to ```--id=A1``` which furtunately is different from sending to ```--id=XA1```. Without the X it would be ambiguous.

This are the technical reasons.

'R' instead of 'X' is also not so bad, because using all the 20 bits as ID is some kind of "raw" ID (R flags raw). But since I choosed 16 different chars for the ID, which are accidentally all the hex digits, the X is a good char to flag this.

This are the philosophical reasons.

CurlyMoo: > The id is a string build from 16 different chars. I could made it a string out of any other number of chars, 8, 19 or 24 or 33 different chars for example. But I choose 16 because this leads to short enough IDs of max. 5 chars.

How can you `choose` a string length. I assume the protocol you are reverse-engineering has a fixed logic, not something you can make various chooses in right?

What you are implementing is just totally different in some aspects than i experiences before, that's why i want to understand what and why...

In the clarus protocol the character is actually an essential part of the raw code being built and is also generated from interpreting the raw code:
https://github.com/pilight/pilight/blob/master/libs/pilight/protocols/433.92/clarus.c#L75-L80
https://github.com/pilight/pilight/blob/master/libs/pilight/protocols/433.92/clarus.c#L141-L157

CMakeConfig.txt
```diff
@@ -80,6 +80,7 @@ set(PROTOCOL_SUNRISESET ON CACHE BOOL "support for the Sunrise / Sunset calculat
 set(PROTOCOL_TECHLICO ON CACHE BOOL "support for TechLiCo Lamps protocol")
 set(PROTOCOL_TEKNIHALL ON CACHE BOOL "support for the Teknihall weather station protocol")
 set(PROTOCOL_TFA ON CACHE BOOL "support for the TFA weather station protocol")
+set(PROTOCOL_UNITEC ON CACHE BOOL "support for the unitec switch protocol")
```
woutput: Shouldn't PROTOCOL_UNITEC be PROTOCOL_UNITEC_SWITCH?
Since I understand that *Unitec* is the brand name and *switch* is the type of device. I believe there should be a difference between device types, since apparently there is also a [push button switch] (https://www.amazon.co.uk/Unitec-47195-Button-Switch-Radio/dp/B00E625IE6?ie=UTF8&*Version*=1&*entries*=0)

libs/pilight/protocols/433.92/unitec.c
```diff
@@ -0,0 +1,249 @@
...
+	if(!end || end==id+1 || *end) {
+		logprintf(LOG_ERR, "unitec_switch: invalid char in id: %s ", end ? end : "?");
+		return EXIT_FAILURE;
+	}
+	if(end -(id+1) > 5) {
+		logprintf(LOG_ERR, "unitec_switch: too long id. Max 5 digits allowed.");
+		return EXIT_FAILURE;
+	}
+
+	if(unit < 0 || unit > 4) {
```
woutput: What is the range of *unit*?
This check allows 0...4, but text on line 191 says 1...4
In addition, regex on line 230 allows 0...5

wo-rasp:
countof() throws an error:
libpilight.a(unitec.c.o): In function `parseCode':
/home/pi/pilight/libs/pilight/protocols/433.92/unitec.c:77: undefined reference to `countof'

libs/pilight/protocols/433.92/unitec.c
yablacky: So why can I do things I did:
I found out that the id is stored in twenty bits in the pulse. By nature these twenty bits are the binary representation of the id.

In a command line or user interface we could use the same (nature) representation of the id. This would lead to names of 20 characters length consisting of 2 different chars, namely digits "0" and "1" (a base-2 number of up to 20 digits).

The 20-bit-id could also be represented by

* decimal numbers from 0 to 1048575 (a base-10 number of up to 7 digits)
* hex numbers from 0 to FFFFF (a base-16 number of up to 5 digits)
* names build out of 26 different chars A .. Z (a base-26 number of up to 5 digits)
* names build out of 36 different chars 0..9 and A .. Z (a base-36 number of up to 4 digits)
* In fact any base could be used unless there are suitable characters to represent a digit.

The best representation of couse is how unitec want the id to be represented. If only we had this information. I don't have it. And the unitec device I used for reverse engineering does not give any hint how an id should look like.

Despite convenience all representations have the same right being used. So I was free to choose a representation and I selected the hex-digit representation. The "X" in front of the id is artificiall added by me to indicate how the remaining characters have to be interpreted (namely as hex digits). One could additionally implement to specify the id in further representations, say decimal numbers. In this case an artificial leading "D" would be suitable to indicate, that the the following characters are decimal digits. Then ``--id XA`` and ``--id D10`` are equivalent and refer the same 20-bit id.

Should, in the future, one can tell the preferred unitec native representation of ids, then it can be added to the implementation without breaking existing configurations which are using ids with leading X. The native ids are just a further representation of the same 20-bits in the pulse like the D-prefix example above.

Don't hesitate to ask for further explanations if needed.

libs/pilight/protocols/433.92/unitec.c
yablacky: The range of unit is 0 .. 4.

Text in line 191 says 1..4 **or 0 for all** which is 0..4 as checked in line 190.

The regex in line 230 is a typo. Will fix that, thank you.

CMakeConfig.txt
yablacky: I can follow your argumentation and even like being precise; but then a lot of protocol names should be changed to be more specific just in case a different kind of device comes under the same brand.

For symmetry of protocol names in pilight I would prefer to wait until one implements the unitec push button and founds out that is really a different protocol.

yablacky:
countof() is now defined in libs/pilight/core/common.h

libs/pilight/protocols/433.92/unitec.c
```diff
@@ -0,0 +1,249 @@
...
+	unitec_switch->message = json_mkobject();
+	json_append_member(unitec_switch->message, "id", json_mknumber(id, 0));
+	json_append_member(unitec_switch->message, "unit", json_mkstring(unit));
+	json_append_member(unitec_switch->message, "state", json_mkstring(state ? "on" : "off"));
+}
+
+static void parseCode(void) {
+	int frame[RAW_LENGTH/2], x = 0;
+
+	if(unitec_switch->rawlen != RAW_LENGTH)
```
CurlyMoo: This should never happen since the validate function also checks for this.

libs/pilight/protocols/433.92/unitec.c
```diff
@@ -0,0 +1,249 @@
...
+	 * https://sites.google.com/site/arduinohomeautomation/home/unitec_eim826
+	 * (there 0 and 1 is vice versa, frame bit positions are counted reverse).
+	 */
+
+	for(x=0; x < countof(frame); x++) {
+		// High "1" if first > second, Low "0" else
+		frame[x] = unitec_switch->raw[x*2] > unitec_switch->raw[x*2+1];
+	}
+
+        int unitn = binToDec(frame, 0, 19);
```
CurlyMoo: Indentation

PR #270, opened 2016-05-06
Oregon Protocol V2.1 support

wo-rasp:
The Oregon Family has three major Protocol Variants V1, V2 and V3.
V2 is the most popular one.
This release supports protocol V2.1, for the supported devices check the pilight wiki.
The development history is discribed in the pilight forum:
https://forum.pilight.org/Thread-Oregon-THGR122NX?pid=1050#pid1050
Several users have tested this protocol based on pilight V6.


CurlyMoo:
First of all, three commits instead of one.

wo-rasp:
Squashed.

CurlyMoo:
Can you go through the changes and their function one by one. You suggest quit some changes in de daemon so that does need some explanation.

wo-rasp:
Okay,
As introduction:
https://forum.pilight.org/Thread-Dummy-protocol-for-master-in-adhoc-network?pid=18086#pid18086

The main modification starts at L1586 and it is basically a state-machine doing some preprocessing of received raw data in order to convert a headerless and footerless protocol into a footer based protocol.

WAIT_FOR_END_OF_HEADER
The oregon protocol has a preamble pulse sequence, and here we search for it, this is required as we have to synchronize the decoder. 

Only pulses with a duration within the range of 732 and 1120 are analysed.
If we find a total of 23 to 32 consecutive pulses with this length, immediately followed by a qualified SYNC pulse, we assume that we have found a valid preamble sequence of oregon V2.1 protocol and we reset the buffer pointer to ensure that the complete payload fits into it and the state machine is now set for preprocessing of payload data: WAIT_FOR_END_OF_DATA.

If we do not find 23 to 32 consecutive pulses immediately followed by a qualified SYNC pulse, no change to raw data or raw buffer handling occurs.

WAIT_FOR_END_OF_DATA
The payload is transmitted twice and either terminated by a GAP of 11018µS or immediately followed by a 2nd pulsetrain. In the latter case the preamble pulses are converted into an artificial footer with the length of 11018.
We cut off the preamb pulses from the received raw data stream, the first pulse is the SYNC pulse, thus the oregon decoder has a known state.
As a result we have now a buffer filled with the 1st payload frame, terminated with a footer of 11018 and the statemachine is set to: WAIT_FOR_END_OF_DATA_2

WAIT_FOR_END_OF_DATA_2
Reception of the 2nd pulsetrain starts either with a received footer pulse or with a preamb pulse. In real life pilight has missed some pulses, thus for the preprocessing of the 2nd pulsetrain parts of the pulses may have to be rebuilt. More information can be found here:
https://forum.pilight.org/Thread-Oregon-THGR122NX?pid=10176#pid10176.
Once the preamb pulses are identified the statemachine enters state WAIT_FOR_END_OF_DATA_3.

WAIT_FOR_END_OF_DATA_3
The 2nd pulsetrain does not end with a qualifying footer but with a pause of unknown length, so we create an 11018 footer pulse and reset the statemachine to WAIT_FOR_END_OF_HEADER.

CurlyMoo:
So what basically happens is that we analyze the stream while it's being received to see if it's Oregon? If so, i strongly believe we can make it more modular.

wo-rasp:
The main functionality is all in one block. Do you have anything specific in mind for more modular and efficient implementation?

I consider the CPU overhead as neglectable at this point, compared to the overhead of validating all buffers received with all protocols defined. (One of the links in the commit message discusses the test experience of dropped pulses after forwarding a filled buffer, and a strategie to recover the missing CLOCK pulses). However in particular on a Pi3 the assignment of available cores to time critical tasks is worth a discussion. 

No, i am not checking for Oregon in particular, i am checking for a sequence of CLOCK pulses at the beginning of a raw buffer, followed by a SYNC pulse. The current implementation is OREGON specific, but as said earlier, SOMFY has the same structure (4 CLOCK pulses followed by a SYNC pulse for the first pulsetrain and 14 CLOCK pulses followed by a SYNC pulse for the repetitive pulsetrains).

If i detect such a condition I reset the buffer pointer in order to avoid a buffer overflow and start saving the incoming raw pulses. Non manchester encoded pulsetrains should not be affected, as the duration of pulses differs sufficiently. There is a slight danger of disrupting other manchester encoded pulsetrains as their payload may qualify as a sequence of CLOCK pulses but I am currently not aware of any such pulsetrain within pilight.

CurlyMoo:
> The main functionality is all in one block. Do you have anything specific in mind for more modular and efficient implementation?

Ofc, for a start, i don't want to have a specific implementation in the main daemon.

> I consider the CPU overhead as neglectable at this point, compared to the overhead of validating all buffers received with all protocols defined. (One of the links in the commit message discusses the test experience of dropped pulses after forwarding a filled buffer, and a strategie to recover the missing CLOCK pulses). However in particular on a Pi3 the assignment of available cores to time critical tasks is worth a discussion. 

Agreed. We don't want full buffers evaluated.

My implementation would be:
1. An additional protocol parameter called `inline` or such.
2. Inline protocols will have a function that is called after each received pulse. The specific protocol function can then evaluate and buffer the raw pulse streams inline.
3. As soon as two sync pulses are received those protocols can immediatly decode the raw pulses instead of having to edit the raw stream with simulated footers.

So let's start by extracting a raw stream. Then tell me (in pseudocode) what should happen. Then i will come up with a more modular solution.

wo-rasp:
Raw Streams:
Both RAW Streams are manchester encoded, thus no definite raw length

Type 1 s_mfy-RTS:
preAmb Code
payload
postAmbCode (various Footers)

Type 2 Oregon Protocol 2.1:
PreAMB SYNC Pulses (Length and duration differs for V1, V2, V3 protocol)
payload
Fixed GAP duration or PreAMB SYNC Pulses (Length and duration differs for V1, V2, V3 protocol
payload
(no further transmission, no further repetition of payload, pause of indefinite length, max # of pulses 460)

wo-rasp:
I think it is time to move on and get this pending issue resolved, I propose an integration in several steps, which will include s_mfy as well. I will setup a new thread in the core program section.

PR #272, opened 2016-05-15
Fix syscodetype 0001 for GT9000 protocol

dominikkarall:
Syscodetype 0001 was using the wrong hash map.


CurlyMoo:
Can you make it a single commit?

dominikkarall:
I will make a single commit as soon as I have a console available. Unfortunately the webinterface of github doesn't support rebase.

PR #277, opened 2016-05-26
TCM 218943 weather station support

puuu:


CurlyMoo:
Is it just me or was there never a wiki entry? I want to include this protocol in the new stable version, but i need a manual page (instead of the deprecated wiki) for it.

puuu:
Sorry, there is was never any wiki page or pilight documentation for this weather station.

The protocol information are basing on a discussion on the arduino forum: http://forum.arduino.cc/index.php?topic=136836.0 (unfortunitly in geman)

A summary can be found as a comment in the [tcm.c](https://github.com/puuu/pilight/blob/development/libs/pilight/protocols/433.92/tcm.c#L40) file.

If you need more information, I may can provide them.

CurlyMoo:
> If you need more information, I may can provide them.

Yes, i need a manual page like this http://manual.pilight.org/protocols/433.92/weather/conrad.html.
The source can be found here: https://github.com/pilight/pilight/blob/development/docs/source/protocols/433.92/weather/_conrad.rst

PR #278, opened 2016-05-27
MaxiTronic switch support

puuu:
Protocol is similar to elro_800_switch, but has a additional Med bit in the ID part.

Transmitter and switch has no dip switches.


CurlyMoo:
Can't we make elro_800_switch more versatile?

puuu:
Did you have any suggestion? I don't know if it is a good idea to change `systemcode` from number to string.

CurlyMoo:
So let's start with a forum post describing the protocol. I now have to guess what you implemented and that not something i do best.

puuu:
Close, because PR #282 is merged.

PR #279, opened 2016-05-27
Action switch support

puuu:


CurlyMoo:
Is action the same as impuls?

puuu:
Hmm, it looks like this. I did not realized before. Indeed, the switch works with impuls, but not the transmitter. 

The average timing from the transmitter is:
short: 190 (175-201)
long: 553 (544-560)
footer: 5798 (5518-5854)

So, we need at least MIN_PULSE_LENGTH=160 in impuls to work. If you think it is better to change impuls please do, or let me know to, if I should make a new pull request.

CurlyMoo:
Impuls was one of the first protocol supported and the last years we haven't had complains of it not working.

puuu:
Ok, but increasing `MIN_PULSE_LENGTH` from 130 to 160 will not break anything? And at least, my remote is working...

CurlyMoo:
So discuss it on the forum and hear if others can confirm.

PR #282, opened 2016-05-29
elro_800_switch: support of Med-states in systemcode

puuu:
Some switches, like the Maxi-Tronic swtich (see pull request #278), using additional Med states (long-short-long-short pulse) in the systemcode. This commit support this kind of switches by supporting systemcodes > 31.

### DIFF
--- a/libs/pilight/protocols/433.92/elro_800_switch.c
+++ b/libs/pilight/protocols/433.92/elro_800_switch.c
@@ -59,7 +59,7 @@ static void createMessage(int systemcode, int unitcode, int state) {
 }
 
 static void parseCode(void) {
-	int binary[RAW_LENGTH/4], x = 0, i = 0;
+	int binary[RAW_LENGTH/4], x = 0;
 
 	if(elro_800_switch->rawlen>RAW_LENGTH) {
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Why is there a new PR instead of working further on the previous one?

CurlyMoo:
So using systemcodes smaller than 31 uses the regular elro logic, above uses the new med approach?

puuu:
Correct, as written on the forum, I have tested it with Elro 800 switches. Remote and switches working like before.

puuu:
btw, I prefer this approach over #278. It is more universal (supporting more than one med) and we have not to deal with an additional protocol (the Maxi-Tronic remote is already recognised by 5 protocols).

puuu:
What can we do to move on with this pull request?
Since several monthes this runs without problems for me and I have many elro800 switches and these Maxi-Tronic switches.

In the [forum](https://forum.pilight.org/Thread-Maxi-Tronic-switch) there is no further discussion about this.

I would prefere to close #278 and merge this pull request.

puuu:
Ok, I was a little bit lazy because all med bits on is  992 so I thought, 999 is an easy and save regex. But now I changed to 1023.

Thanks

PR #284, opened 2016-06-04
GT9000: Change short pulse to 500

dominikkarall:
Some sockets had problems with 300. 500 should work for all of them.


CurlyMoo:
Version number?

dominikkarall:
Version updated to 1.3

PR #286, opened 2016-06-13
Fix blank warning log messages

pilino1234:
These occurred for example when accessing an invalid page on the webserver (x.x.x.x:5000/conf for example)
Before:
```
[Jun 13 12:55:22:359922] WARNING:[Jun 13 12:55:22:567092] DEBUG: new client, ip: 127.0.0.1, port: 51047
```

After:
```
[Jun 13 13:19:55:969887] WARNING: (webserver) could not read /usr/local/share/pilight//conf
```

### DIFF
--- a/libs/pilight/core/log.c
+++ b/libs/pilight/core/log.c
@@ -142,11 +142,11 @@ void logprintf(int prio, const char *str, ...) {
 		
 		/* len + loglevel */
 		if(len+9 > bufsize) {
-			if((buffer = realloc(buffer, len+9+1)) == NULL) {
+			if((buffer = realloc(buffer, len+10)) == NULL) {
 				printf("out of memory\n");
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Why should the fix only apply to LOG_WARNING?

pilino1234:
It seems that the "WARNING: " string is too long (9 characters), removing any one or more character(s) from it fixes the truncated log message (for example changing it to "WARNIN: "). Increasing the size by 1 fixes this.

CurlyMoo:
I think instead the `len+9` should be `len+10` as well as `bufsize = len+9+1` to `bufsize = len+10+1`? Can you check that?

pilino1234:
Yup, that works too. I'll update the commit.

PR #289, opened 2016-07-01
STEP 1 Modification of daemon.c

wo-rasp:
As discussed here:
https://forum.pilight.org/Thread-Integration-of-PR-270-oregon-and-269-s-mfy-branch-into-pilight-V7?pid=18636#pid18636
this is the first Step to enclose the current daemon code for header/footer detection in a separate function.
In a second step a new user function will be added for enhanced header/footer detection.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -1499,36 +1499,41 @@ void *receivePulseTrain(void *param) {
 	return (void *)NULL;
 }
 
-int sys_parseHeader(struct hardware_t *hw, struct rawcode_t r, struct timeval tp, struct timespec ts, int duration) {
+int usr_parseHeader(struct hardware_t *hw, struct rawcode_t *r, struct timeval *tp, struct timespec *ts, int *duration) {
+return -1;
+}
+
+
+int sys_parseHeader(struct hardware_t *hw, struct rawcode_t *r, struct timeval *tp, struct timespec *ts, int *duration) {
 int plslen = 0;
 
-	if(duration > 0) {
-		r.pulses[r.length] = duration;
-		r.length++;
-		if(r.length > MAXPULSESTREAMLENGTH-1) {
-			r.length = 0;
+	if(*duration > 0) {
+		r->pulses[r->length] = *duration;
+		r->length++;
+		if(r->length > MAXPULSESTREAMLENGTH-1) {
+			r->length = 0;
 		}
-		if(duration > hw->mingaplen) {
-			if(duration < hw->maxgaplen) {
-				plslen = duration/PULSE_DIV;
+		if(*duration > hw->mingaplen) {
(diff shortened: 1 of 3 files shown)


CurlyMoo:
Already said i have a different approach in mind...

wo-rasp:
That is understood, however as said none of the OREGON devices has a predictable footer nor header pulses, some devices insert a known GAP of 1100 µS between the repetition of the payload, while others repeat the SYNC pulse sequence. The duration of pulses is between 350 and 1100µS, depending on the RF state and the binary state.

To detect an OREGON pulsetrain, ii is required to analyse each incoming pulsestream for pulse patterns with specific characteristics, in this particular case for a pattern of 32 pre-amb pulses followed by 12 SYNC pulses followed by an unknown number of data pulses representing the payload with an unknown binary length.
All three groups of these pulses are manchester encoded and have the following specific characteristics: The 32 pre-amb pulses are all equal and the last one represents a logical 1 condition, the SYNC pulses are presenting logically the binary value 10101010, thus the first two pulse of the SYNC sequence are two short pulses followed by a long one. In order to synchronize your decoder is is mandatory to detect this series of 4 (four) qualifying SYNC pulses followed by the data pulses. Each data bit is always sent twice (inverted - non-inverted). The entire pulsetrain is sent twice, omitting the pre-amb pulses, some sensors will insert a gap (of appr. 11000µS) while others will sent the SYNC pulse sequence instead. There is no footer, transmission simply ends.

I do consider as design parameter for a Pi a CPU power of appr. 100MIPS and for a Pi2 of 1000MIPS.
If we want to grab pulses with an accuracy of 25µS every 350µS and want to keep CPU load below 10% I propose to keep overall runtime in the range of 20 to 50µS for a Pi, i estimate that we should try to stay with our code below 2000 CPU instructions. Using a Pi2 or Pi3 allows more CPU instructions, but we still do need to keep an eye on thread overhead and data buffer handling, as some oregon devices will make use of the maximum buffer size of 512 byte (I do need to calculate that in a more precise manner, as a 1st estimate ended up with a worst case scenario of 540 byte.), and each individual parser would require its own raw data buffer.

We also do need to keep an eye on users without LPF or operation in a noisy environment with interrupts coming in in less than 100µS intervals.

In brief, I am not convinced that a universal approach of adding an individual parser for each protocol will carry us any further than supporting more than two or three user defined protocol implementations on a PI architecture, thus i propose to go ahead with the current approach and gain experience with the current implementation in V7 and obtain feedback from other users.

I agree that another approach, making full use of the Multi Core capabilities of the Pi3 architecture, is possible.

CurlyMoo:
I think my idea will fit your requirements while also allowing a more modular approach, but i need a raw pulsetrain to present it.

PR #304, opened 2016-11-22
modified daemon.c for receive-repeats rxrpt per protocol

frenchie71:
this adds a check for protocol->rxrpt to daemon.c Purpose: avoid false alarms with desired protocols and enable us to set receive-repeats on a protocol level. In some configurations (e.g. smoke detectors) it is desirable to add a minimum of receive repeats before triggering an alarm. If you modify rxrpt in your protocol to a value n>1, the first n-1 receives will be discarded.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -606,10 +606,15 @@ void *receive_parse_code(void *param) {
 						protocol->repeats++;
 						if(protocol->parseCode != NULL) {
 							logprintf(LOG_DEBUG, "recevied pulse length of %d", recvqueue->plslen);
-							logprintf(LOG_DEBUG, "caught minimum # of repeats %d of %s", protocol->repeats, protocol->id);
 							logprintf(LOG_DEBUG, "called %s parseRaw()", protocol->id);
 							protocol->parseCode();
-							receiver_create_message(protocol);
+							if (protocol->repeats < protocol->rxrpt ) {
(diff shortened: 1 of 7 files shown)


CurlyMoo:
The rxrpt is not used anywhere at the moment so i'm open for integrating this as soon as it's being used somewhere.

frenchie71:
I think people are using it in specific implementations. I am using it hard coded for the protocols which I have trouble with, but this is most certainly nothing that should be pushed to a generic repository but which we may discuss in the forum. The bottom line is, that people _may_ use rxrpt together with this change. Alternatively there is no point in having an rxrpt variable on the protocol object. The change is done in a way so that it is 100% transparent to existing implementations. The long term plan would probably be to open protocols for configuration options in config.json

CurlyMoo:
> The bottom line is, that people may use rxrpt together with this change. 

True, but those who do also easily find this PR :smile:

frenchie71:
Well, I do partially agree ;-) It is true that people may find this PR but on the other hand they will most probably not google for rxrpt or similar but rather for "smoke detector false alarms" etc. Anyhow, it's your project and therefore your call. I'll open a thread on the forum and describe what I had in mind when I did this - after all I understand that you are sceptical because I have not yet contributed to the project and am showing up out of the blue.

CurlyMoo:
Let's first do the topic and then we'll see.

PR #306, opened 2016-11-30
Dimmer patches

pilino1234:
Fix 2 dimmer bugs

These are two separate commits since they are quite different, yet I feel that they are too similar to warrant opening separate PRs. Hope that's fine.

Discussion: https://forum.pilight.org/Thread-Dimlevel-text-in-WebGUI-0-and-not-updated

Commit messages: 

----
Webgui: Fix dimmer display bugs

This would cause the text field next to the slider to be
initialized to 0 at page load, even though the slider was showing the
correct dimlevel.

This commit fixes this by also assigning the dimlevel to the text field on
page load. Patch by @Niek.

Further, the slider-input field was editable, but without use (editing the
value has no effect). Therefore, display the field as disabled.

----

generic_dimmer: Fix inconsistency

The default max values were defined to different values in the checkValues and
createCode functions, causing the webgui to not transmit slider changes to
values above 10.
This commit fixes this by setting the default to 15 in both functions.

### DIFF
--- a/libs/pilight/protocols/generic/generic_dimmer.c
+++ b/libs/pilight/protocols/generic/generic_dimmer.c
@@ -30,6 +30,9 @@
 #include "../../core/gc.h"
 #include "generic_dimmer.h"
 
+#define MIN_DIMMER_VALUE 0
+#define MAX_DIMMER_VALUE 15
+
 static void createMessage(int id, int state, int dimlevel) {
 	generic_dimmer->message = json_mkobject();
 	json_append_member(generic_dimmer->message, "id", json_mknumber(id, 0));
@@ -46,8 +49,8 @@ static void createMessage(int id, int state, int dimlevel) {
 
 static int checkValues(JsonNode *code) {
 	int dimlevel = -1;
-	int max = 15;
-	int min = 0;
+	int max = MAX_DIMMER_VALUE;
+	int min = MIN_DIMMER_VALUE;
 	double itmp = -1;
 
 	if(json_find_number(code, "dimlevel-maximum", &itmp) == 0)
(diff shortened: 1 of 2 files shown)


libs/pilight/protocols/generic/generic_dimmer.c
```diff
@@ -75,7 +75,7 @@ static int createCode(JsonNode *code) {
 	int id = -1;
 	int state = -1;
 	int dimlevel = -1;
-	int max = 10;
+	int max = 15;
```
CurlyMoo: Better make this a global `define`

pilino1234:
Done.

libs/pilight/protocols/generic/generic_dimmer.c
```diff
@@ -75,7 +77,7 @@ static int createCode(JsonNode *code) {
 	int id = -1;
 	int state = -1;
 	int dimlevel = -1;
-	int max = 10;
+	int max = MAX_DIMMER_VALUE;
 	int min = 0;
```
CurlyMoo: The same counts of course for the minimal dimlevel :smile:

pilino1234: Hmm, forgot about that. Done :)

CurlyMoo:
Thanks

CurlyMoo:
You forgot the webgui version increase. Stupid me.

pilino1234:
How to go about resolving this? I've amended the commit in by branch. I think the only way is to open a new PR since this one can't be updated any longer.

CurlyMoo:
I know. A new PR is the only way.

pilino1234:
What should the new PR include? Should the commits from this PR be reverted and added to a new PR, or just a commit to bump the webgui version?

CurlyMoo:
Just the bump

pilino1234:
Will do

PR #309, opened 2016-11-30
TCM 218943 weather station: support negative temperature

puuu:
The temperature is a signed integer, so we have to handle the negative case.

### DIFF
--- a/libs/pilight/protocols/433.92/tcm.c
+++ b/libs/pilight/protocols/433.92/tcm.c
@@ -43,7 +43,7 @@
 	9	low battery indicator
 	12	1 indicates TX button press on sensor
 	17-24	humidity
-	25-36	temperature
+	25-36	temperature (signed int12)
 */
 
 typedef struct settings_t {
@@ -93,6 +93,9 @@ static void parseCode(void) {
 	humidity = binToDecRev(binary, 16, 23);
 
 	temp = binToDecRev(binary, 24, 35);
+	if(binary[24]) {
+		temp -= 4096;
+	}
 	temperature = temp;
 
 	struct settings_t *tmp = settings;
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Has this been tested?

dc6jn:
Just came across this topic. i found 2 weeks ago a bug that might be related.
In the driver protocols/433.92/alecto_wx500.c there is also a bug connected with negative temperature values. The problem was solved by me with the following patch:
--------------------------------------------
	json_append_member(alecto_wx500->message, "data", json_mknumber(data,12));
	switch(type) {
		case 1:
			id = binToDec(binary, 0, 7);
			long x=binToDec(binary, 12, 23);
                        //the value is a  temperature 
                        //This 12 bits wide 2's complement signed binary number represents the actual temperature value in 0.1 °C units.
			x = (x >> 11) == 0 ? x : -1 ^ 0xFFF | x;
			temperature = (double)x/10.0;
			humidity = (binToDec(binary, 28, 31) * 10) + binToDec(binary, 24,27);
-------------------------------------------------------------------
Hope this helps

puuu:
@CurlyMoo: Yes, I have three of these transmitter. One of them is outdoor and measured last night down to -3°C. With this patch, the displayed temperature (Display of the transmitter) and the parsed value are the same, for all of the three transmitter. Before the patch, the received and decoded value was about 407°C at negative temperatures.

puuu:
@dc6jn I think this was fixed in #234.

@CurlyMoo It seems, there are many device/protocols that use signed integer (e.g. #234, #221).  IMHO we should consider to introduce a function to convert binary to signed integer.

e.g:
```C
int binToSigned(const int *binary, int s, int e) { //  0<=s<=e, binary[s(lsb) .. e(msb)]
	int result;
	result = binToDec(binary, s, e);
	if(binary[s]) {
		result -= 1<<(e-s+1);
	}
	return result;
}
```

CurlyMoo:
> IMHO we should consider to introduce a function to convert binary to signed integer.

Good idea (in a new PR).

PR #311, opened 2016-12-01
Support for the BH1750 i2c illuminance sensor

GreenMarc:
Here is the support for the bh1750 i2c illuminance sensor. See the following forum thread : https://forum.pilight.org/Thread-BH1750-i2c-digital-illuminance-sensor

This code is based from the lm75 i2c temperature sensor.

Greetings
Marc


pilino1234:
Same here, can you squash these three commits into one please?

CurlyMoo:
Can you create a unit test for it as already can be found for some i2c devices in the rewrite branch?

PR #312, opened 2016-12-01
Support for the TSL2561 i2c illuminance sensor

GreenMarc:
Here is the support for the tsl2561 i2c illuminance sensor. See the following forum thread : https://forum.pilight.org/Thread-TSL2561-i2c-digital-illuminance-sensor

This code is based from the lm75 i2c temperature sensor.

Greetings
Marc


CurlyMoo:
Never close PR's.

PR #313, opened 2016-12-01
Support for the TSL2561 i2c illuminance sensor

GreenMarc:
Here is the support for the tsl2561 i2c illuminance sensor. See the following forum thread : https://forum.pilight.org/Thread-TSL2561-i2c-digital-illuminance-sensor

This code is based from the lm75 i2c temperature sensor and the sample code in the datasheet of the tsl2561

Greetings
Marc


pilino1234:
Can you squash these three commits into one (as in the contributing.md) please? This avoids cluttering the commit history of the repository. A link to instructions on how to do this is in the contributing.md file.

PR #322, opened 2016-12-23
Add schema export action for protocols

janLo:
This adds a new action called "protocol_schema" to the daemon, to
export the send-schema for all supported protocols to a client. This
can be used to build client-side data validation for better error
propagation and/or mor stable api implementations.

Signed-off-by: Jan Losinski <losinski@wh2.tu-dresden.de>

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -1265,6 +1265,23 @@ static void socket_parse_data(int i, char *buffer) {
 					} else {
 						socket_write(sd, "{\"status\":\"failed\"}");
 					}
+				} else if(strcmp(action, "list protocols") == 0) {
+					struct protocols_t *pnode = protocols;
+					struct JsonNode *schema = json_mkobject();
+					struct JsonNode *proto_schema = json_mkarray();
+
+					json_append_member(schema, "protocols", proto_schema);
+
+					/* iterate all protocols */
+					while(pnode) {
+						json_append_element(proto_schema, protocol_schema(pnode->listener));
+						pnode = pnode->next;
+					}
+
+					char *schema_message = json_stringify(schema, NULL);
+					socket_write(sd, schema_message);
+					json_free(schema_message);
+					json_delete(schema);
 				} else if(strcmp(action, "control") == 0) {
 					struct JsonNode *code = NULL;
 					struct devices_t *dev = NULL;
--- a/libs/pilight/core/json.c
+++ b/libs/pilight/core/json.c
@@ -1436,3 +1436,26 @@ int json_find_string(JsonNode *object, const char *name, char **out) {
 void json_free(void *a) {
 	free(a);
 }
+
+JsonNode *json_vartype(int vartype) {
+	JsonNode *vartype_array = json_mkarray();
+	if((vartype & JSON_NULL) != 0) {
+		json_append_element(vartype_array, json_mkstring("null"));
+	}
+	if((vartype & JSON_BOOL) != 0) {
+		json_append_element(vartype_array, json_mkstring("boolean"));
+	}
+	if((vartype & JSON_STRING) != 0) {
+		json_append_element(vartype_array, json_mkstring("string"));
+	}
+	if((vartype & JSON_NUMBER) != 0) {
+		json_append_element(vartype_array, json_mkstring("number"));
+	}
+	if((vartype & JSON_ARRAY) != 0) {
+		json_append_element(vartype_array, json_mkstring("array"));
+	}
+	if((vartype & JSON_OBJECT) != 0) {
+		json_append_element(vartype_array, json_mkstring("object"));
+	}
+	return vartype_array;
+}
--- a/libs/pilight/core/json.h
+++ b/libs/pilight/core/json.h
@@ -121,4 +121,6 @@ bool json_check(const JsonNode *node, char errmsg[256]);
 int json_find_number(JsonNode *object, const char *name, double *out);
 int json_find_string(JsonNode *object, const char *name, char **out);
 
+JsonNode *json_vartype(int vartype);
+
 #endif
--- a/libs/pilight/protocols/protocol.c
+++ b/libs/pilight/protocols/protocol.c
@@ -347,6 +347,39 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
 	proto->devices = dnode;
 }
 
+JsonNode *protocol_schema(protocol_t *proto) {
+	JsonNode *root = json_mkobject();
+
+	json_append_member(root, "name", json_mkstring(proto->id));
+
+	JsonNode *devices = json_mkarray();
+	json_append_member(root, "devices", devices);
+
+	struct protocol_devices_t *dev_temp = proto->devices;
+	while(dev_temp) {
+		json_append_element(devices, json_mkstring(dev_temp->id));
+		dev_temp = dev_temp->next;
+	}
+
+
+	JsonNode *options = json_mkarray();
+	json_append_member(root, "options", options);
+
+	struct options_t *opt_temp = proto->options;
+	while(opt_temp) {
+		if(opt_temp->id != 0) {
+			JsonNode *option = json_mkobject();
+			json_append_element(options, option);
+
+			json_append_member(option, "name", json_mkstring(opt_temp->name));
+			json_append_member(option, "vartype", json_vartype(opt_temp->vartype));
+		}
+		opt_temp = opt_temp->next;
+	}
+
+	return root;
+}
+
 int protocol_device_exists(protocol_t *proto, const char *id) {
 	logprintf(LOG_STACK, "%s(...)", __FUNCTION__);
 
--- a/libs/pilight/protocols/protocol.h
+++ b/libs/pilight/protocols/protocol.h
@@ -128,6 +128,7 @@ void protocol_set_id(protocol_t *proto, const char *id);
 void protocol_plslen_add(protocol_t *proto, int plslen);
 void protocol_register(protocol_t **proto);
 void protocol_device_add(protocol_t *proto, const char *id, const char *desc);
+JsonNode *protocol_schema(protocol_t *proto);
 int protocol_device_exists(protocol_t *proto, const char *id);
 int protocol_gc(void);
 


daemon.c
```diff
@@ -1245,6 +1245,24 @@ static void socket_parse_data(int i, char *buffer) {
...
+
+					json_append_member(schema, "protocols", proto_schema);
+
+					/* iterate all protocols */
+					while(pnode) {
+						json_append_element(proto_schema, protocol_schema(pnode->listener));
+						pnode = pnode->next;
+					}
+
+					char *schema_message = json_stringify(schema, "\t");
```
CurlyMoo: Why are you sending formatted message over socket.

janLo:
rebased on top of development ...

@CurlyMoo do I just set the second parameter to NULL to skip formatting?

janLo:
Ok. Tested, changed, PR updated :)

CurlyMoo:
Any documentation on this?

janLo:
Where shall the documentation go?

CurlyMoo:
First here on github to convince me of merging this :smile:

janLo:
Ah, ok. Here we go:

Its a new "action" for the daemon. At the moment called `protocol_schema` (I din't found a better name for it). It returns a json document with the full listing of protocols, the devices they provide and the options they take in the `send` action.

The purpose is, to validate data in third-party-clients before its send to the daemon. The main reason was a discussion in the Homeassistant-Project about how detailed the data should be verified before its send over the socket. The maintainer of the HA-pilight component thinks, it's not good to start to build static defined validation schemes in the HA-software, because this might cause issues when pilight changes the api or implements new protocols. 

On the other hand, the daemon is really picky about the data it gets. If anything that should be a number is transferred as a json-string, it just ignores the argument, let the action fail and writes "insufficient number of arguments" to the log. The user has then to guess/read the pilight-source/debug/try, what to do to make it work.

My opinion is, that the pilight daemon  should provide a spec, what it accepts. So the client can decide, if the data it got from the user, matches the spec, can be converted to match or should be rejected (with a proper reason that can be displayed to the user).

The schema-data (protocols/devices/options) has to be extracted of a live daemon, because its declared programmatically in the protocol init functions. So if I don't want to manually extract (and maintain) it or parse the info out of the C-Code (which I actually tried ;) ), making the schema dump an action is the most consistent method.

First of all I'll use this data in the [pilight-python-client ](https://github.com/DavidLP/pilight) to make it available for a software like home-assistant. But I think it can be used in all kinds of clients. Even to build automatic frontends for known devices.

The format of the request-data is simply:
```JSON
{"action": "protocol_schema"}
```
The reply looks like:
```JSON
{"protocols": [{"devices": ["ping"],                                          
                "name": "ping",                                                
                "options": [{"data_type": "string", "id": 105, "name": "ip"},   
                            {"data_type": "string",                             
                             "id": 100,                                      
                             "name": "disconnected"},                          
                            {"data_type": "string",                            
                             "id": 99,                                         
                             "name": "connected"}]}, 
        ...
               {"devices": ["maxitronic", "brennenstuhl", "elro_800_switch"],
                "name": "elro_800_switch",                                   
                "options": [{"data_type": "string", "id": 102, "name": "off"}, 
                            {"data_type": "string", "id": 116, "name": "on"}, 
                            {"data_type": "number",                            
                             "id": 117,                                                                                              
                             "name": "unitcode"},                            
                            {"data_type": "number",                             
                             "id": 115,                                      
                             "name": "systemcode"}]}, 
        ...
               {"devices": ["iboutique", "alecto_ws1700"],                     
                "name": "alecto_ws1700",                                       
                "options": [{"data_type": "number",                            
                             "id": 98,                                          
                             "name": "battery"},                               
                            {"data_type": "number",                           
                             "id": 104,                                         
                             "name": "humidity"},                            
                            {"data_type": "number", "id": 105, "name": "id"},
                            {"data_type": "number",                            
                             "id": 116,                                       
                             "name": "temperature"}]},                          
               {"devices": ["process"],                                      
                "name": "process",                                           
                "options": [{"data_type": "number", "id": 114, "name": "ram"}, 
                            {"data_type": "number", "id": 99, "name": "cpu"}]}]}
```

Thank you.

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,40 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
 	proto->devices = dnode;
 }
 
+JsonNode* protocol_schema(protocol_t *proto) {
```
CurlyMoo: Can you follow our style guide throughout the code in regard to the pointers.

janLo: Done.

libs/pilight/core/json.c
```diff
@@ -1436,3 +1436,21 @@ int json_find_string(JsonNode *object, const char *name, char **out) {
 void json_free(void *a) {
 	free(a);
 }
+
+JsonNode *json_vartype(int vartype) {
+	if (vartype == JSON_NULL) {
+		return json_mkstring("null");
+	} else if (vartype == JSON_BOOL) {
```
CurlyMoo: Style guide, and can you make this a switch?

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,40 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
 	proto->devices = dnode;
 }
 
+JsonNode *protocol_schema(protocol_t *proto) {
+	JsonNode* root = json_mkobject();
+
+	json_append_member(root, "name", json_mkstring(proto->id));
+
+	JsonNode* devices = json_mkarray();
```
CurlyMoo: Style guide.

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,40 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
 	proto->devices = dnode;
 }
 
+JsonNode *protocol_schema(protocol_t *proto) {
+	JsonNode* root = json_mkobject();
```
pilino1234: Here too

libs/pilight/protocols/protocol.c
janLo: Done.

(maybe provide a clang-format config or something similar)

libs/pilight/core/json.c
janLo: Should be alright now?

libs/pilight/protocols/protocol.c
janLo: Done

pilino1234:
I like this feature a lot, came across this issue when working on my python pilight socket API wrapper and other related projects. 
One thing I'm wondering about this the "id" that every option has. Were does this come from? And how could it be used?

janLo:
The id is the "id" that is set for all "commandline" options of the protocols. Its mainly the ascii-value of the short-option (so it would be 117 fo a `-u` shortopt for `--unitcode` for example). You can see it here:https://github.com/pilight/pilight/blob/master/libs/pilight/protocols/433.92/elro_800_switch.c#L204 - the `'u'` is the "id".

pilino1234:
So it's just the ASCII of the short option from pilight-send? Not sure how that could be used... :)  I guess if it doesn't have a use I don't think it should be exposed.

janLo:
It can easily be tripped out. I thought it might be useful for someone who wants to provid the same "options" as `pilight-send` in his implementation.

janLo:
I've vour things I'm not exactly sure about:

* do we need the id? (On the other hand, It might not hurt to have it)
* Are the options that has a id other than 0 assigned really all the needed options?
* Shall we put some hint about required/not required in it and how do we get this information?
* Is `protocol_schema` a good name? ;)

pilino1234:
I think it would be useful enough without the `"id"`, which would also solve the next two points. 

`protocol_listing`, `protocol_roster`, `protocol_overview`. Or something that is more similar to the existing actions (https://manual.pilight.org/en/api#pf3): `"list protocols"`

janLo:
Ok, I'll name it `list protocols` but I don't see, how the removing of the id solves the second and the third question

pilino1234:
Oh, I misread that quite a lot.... 
Regarding the second question: how would a value of 0 occur for an option? Also note that there are some protocols that do not work with certain combinations of options, which will complicate things quite a lot, for example: https://github.com/pilight/pilight/blob/master/libs/pilight/protocols/433.92/arctech_switch.c#L192
So I think it is fine as it is. 
I don't think there's a way to retrieve the required-ness and optional-ness of options either (regarding the third point).

janLo:
I think, there is no id for non-commandline options like this: https://github.com/pilight/pilight/blob/master/libs/pilight/protocols/433.92/elro_800_switch.c#L208-L209

pilino1234:
Those are `GUI_SETTING` options, which are never specified on the commandline, only in the "gui" section of config.json

janLo:
Ok, so it should be fine to just select the options with non-zero ids?

pilino1234:
Yes, those will be the valid commandline options of the protocols. 
Although I'm still not sure whether listing all command line options really is needed. I think listing all protocols, their supported devices, and the possible values (such as id, temperature, state, dimlevel, etc.) 
But if it's easier to leave it like it is, I think it's fine.

janLo:
I my usecase its sufficient to just have the datatypes of the options.

janLo:
So is it possible to integrate this in the next release (and when is this planned) ?

pilino1234:
Once it's been tested a bit more it could probably be merged, it is a useful feature. 

Next release is whenever curlymo has time to finish rewriting the pilight core to make it better in all the ways :)
See the rewrite branch here on GitHub for the progress that he's published so far. Changes and additions from the current development branch (which is still based on the "old" pilight codebase) will be either already included or ported to the rewrite code later.

CurlyMoo:
It would be great if someone can already adapt `pilight-send` and `pilight-control` to use this approach instead of validating against the actual protocols. That would detach the clients even more from the main daemon.

janLo:
I had to replace the switch statement with an if again, because some options define multiple argument types (generic_label for example). So I decided, that the `data_type` will be an array of one or more possible datatypes.

Sadly this mechanism is generally broken in the current pilight, because the type defines does not use values that are distinguishable by bitwise operations. So we have to fix this first by merging PR #323

libs/pilight/core/json.c
```diff
@@ -1436,3 +1436,28 @@ int json_find_string(JsonNode *object, const char *name, char **out) {
 void json_free(void *a) {
 	free(a);
 }
+
+JsonNode *json_vartype(int vartype) {
```
CurlyMoo: I think it would be better to change this to a function that can be used like this:
`json_append_element(vartype_array, json_mkstring(json_vartype(JSON_ARRAY)))`

CurlyMoo:
> So I decided, that the data_type will be an array of one or more possible datatypes.

Can you give an example?

janLo:
Here is what pilight would export at the moment: https://github.com/janLo/pilight/blob/0a47ee3fba05d90378aae4262728aeaf5da0381e/pilight/protocol_list.json

CurlyMoo:
I think `data_type` can be better renamed to `vartype`.

janLo:
Done.

janLo:
What is the status here?

libs/pilight/core/json.c
```diff
@@ -1436,3 +1436,28 @@ int json_find_string(JsonNode *object, const char *name, char **out) {
...
+	JsonNode *vartype_array = json_mkarray();
+
+	if (0 != (vartype & JSON_NULL)) {
+		json_append_element(vartype_array, json_mkstring("null"));
```
CurlyMoo: Can you change this to pilight style guide.

janLo: Yes. Can you point out what exactly or provide a profile for clang-format or a similar tool?

CurlyMoo: Just as you did elsewhere, remove the space between the if and the hook. I also would like to swap the condition:
`(vartype & JSON_NULL) != 0`

janLo: Done.?

libs/pilight/core/json.c
```diff
@@ -1436,3 +1436,26 @@ int json_find_string(JsonNode *object, const char *name, char **out) {
 void json_free(void *a) {
 	free(a);
 }
+
+JsonNode *json_vartype(int vartype) {
+	JsonNode *vartype_array = json_mkarray();
+	if ((vartype & JSON_NULL) != 0) {
```
CurlyMoo: There are still spaces between the IF the hooks.

janLo: What exactly do you mean? I used the spaces exactly as they're used in the rest of the file - or at least cannot see any difference. Can you please provide a profile for a code formatting tool?

pilino1234:
Right after the `if`, there should be no space before the opening
bracket `(`. So change: `if (` to `if(`

libs/pilight/core/json.c
pilino1234: Right after the `if`, there should be no space before the opening
bracket `(`. So change: `if (` to `if(`

janLo: Should be done now :)

daemon.c
```diff
@@ -1265,6 +1265,24 @@ static void socket_parse_data(int i, char *buffer) {
...
+
+					/* iterate all protocols */
+					while(pnode) {
+						json_append_element(proto_schema, protocol_schema(pnode->listener));
+						pnode = pnode->next;
+					}
+
+					char *schema_message = json_stringify(schema, NULL);
+					socket_write(sd, schema_message);
+
```
CurlyMoo: Can you remove this white line?

daemon.c
```diff
@@ -1265,6 +1265,24 @@ static void socket_parse_data(int i, char *buffer) {
...
+					free(schema_message);
```
CurlyMoo: Can you use json_free here?

libs/pilight/core/json.c
CurlyMoo: What about this suggestion?

janLo: No, the problem here is, that json_vartype can give you an array with more than one type. For example have generic_label `JSON_STRING | JSON_NUMBER` as type for the `label` option. This is why I switched this back from being a switch statement. This would require `json_append_element` to add more than one element at once.

daemon.c
janLo: done.

daemon.c
janLo: done.

libs/pilight/core/json.c
CurlyMoo: True, need to look into this a bit further.

pilino1234:
Can you also write up a draft manual paragraph for the pilight manual? See the [pilight/pilight-manual](github.com/pilight/pilight-manual) repo for this. The online manual only describes the current stable, but I think it would be good to already have this when the next version comes around.

janLo:
At least for German and English this should be possible. There are just two questions:
* What is the License of the manual (it only says `please notice that these manuals are not open source licensed` - So I won't just clone it to make a PR as this could be a possible unallowed redistribution)
* How are these `fodt` files created? I guess (or hope) they're not written with a text-editor.

CurlyMoo:
> What is the License of the manual (it only says please notice that these manuals are not open source licensed - So I won't just clone it to make a PR as this could be a possible unallowed redistribution)

It means that i don't want other to just copy what we wrote for pilight without any effort in return (yet).

janLo:
So is it allowed for me to copy (clone) it to make a PR? Maybe put up a CC-(BY)-NC-SA license. So people can copy and use it as long as they don't use it in a commercial context.

CurlyMoo:
I also don't want other to just copy the documentation without my permission. That's the point. I already notice that a look of people copy the protocol implementations of pilight. Just don't want them to copy the documentation as well without something in return.

janLo:
Then you could at least grant people the permission to copy it to make changes and contribute them back. At the moment you put everyone who clones it to contribute on the risk to get legal problems. You also put yourself on risk that contributors complain when you publish their changes because they never agreed on any license.

There are a lot of licenses out - try to find one that suits or write your own (with a simple set of rules). But no license means "stay away".

Anyway. My main questions where:
* can the copyright-owner grant me the permission to clone it public on my github account to make changes to it to contribute them back?
* what software is recommended to edit the files?

pilino1234:
> what software is recommended to edit the files?

LibreOffice, available for basically all the platforms

CurlyMoo:
by-nc-sa it is 👍

janLo:
Is it possible to merge this?

libs/pilight/protocols/protocol.h
```diff
@@ -128,6 +128,7 @@ void protocol_set_id(protocol_t *proto, const char *id);
 void protocol_plslen_add(protocol_t *proto, int plslen);
 void protocol_register(protocol_t **proto);
 void protocol_device_add(protocol_t *proto, const char *id, const char *desc);
+JsonNode *protocol_schema(protocol_t *proto);
```
CurlyMoo: Can you make this `struct JsonNode`?

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,39 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
...
+	json_append_member(root, "devices", devices);
+
+	struct protocol_devices_t *dev_temp = proto->devices;
+	while(dev_temp) {
+		json_append_element(devices, json_mkstring(dev_temp->id));
+		dev_temp = dev_temp->next;
+	}
+
+
+	JsonNode *options = json_mkarray();
```
CurlyMoo: Likewise

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,39 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
 	proto->devices = dnode;
 }
 
+JsonNode *protocol_schema(protocol_t *proto) {
+	JsonNode *root = json_mkobject();
+
+	json_append_member(root, "name", json_mkstring(proto->id));
+
+	JsonNode *devices = json_mkarray();
```
CurlyMoo: Likewise

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,39 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
 	proto->devices = dnode;
 }
 
+JsonNode *protocol_schema(protocol_t *proto) {
```
CurlyMoo: Likewise

libs/pilight/protocols/protocol.c
```diff
@@ -347,6 +347,39 @@ void protocol_device_add(protocol_t *proto, const char *id, const char *desc) {
...
+	JsonNode *root = json_mkobject();
```
CurlyMoo: And likewise

libs/pilight/core/json.h
```diff
@@ -121,4 +121,6 @@ bool json_check(const JsonNode *node, char errmsg[256]);
 int json_find_number(JsonNode *object, const char *name, double *out);
 int json_find_string(JsonNode *object, const char *name, char **out);
 
+JsonNode *json_vartype(int vartype);
```
CurlyMoo: Likewise

libs/pilight/core/json.c
```diff
@@ -1436,3 +1436,26 @@ int json_find_string(JsonNode *object, const char *name, char **out) {
 void json_free(void *a) {
 	free(a);
 }
+
+JsonNode *json_vartype(int vartype) {
+	JsonNode *vartype_array = json_mkarray();
```
CurlyMoo: Likewise

CurlyMoo:
I would like to merge this first in the rewrite code with proper unit testing. We can then backport it to development. Using this approach let's us anticipate on future bugs,

janLo:
I'm waiting for this functionality since December, because I need it to create proper clients. Before this I have no possibility to validate data before I send it to pilight (and the errors pilight gives - if any - does not really point out, what the problem was).

When Do you think will your rewrite be ready? Is there any possibility to at least have a static schema definition that clients can use to verify the data? At the moment, the only thing one can do if data gets rejected is, to print out the data and read the C-Code to guess why it was rejected.

janLo:
Did these changes land somewhere in the code now?

CurlyMoo:
No, because that unittest was never made. Maybe i was unclear at that point, but no new code has been acccepted in pilight since a long time when it isn't unit tested. So, if you can fix that, i will keep my promise.

janLo:
Well, that's sad. I wrote this code 2 years ago for a version of pilight without any testing infrastructure. Then you delayed the merge over and over again, commingled up with new requests after every ping I send. For me, I've migrated long away from pilight partly because of the unreliability of the 433mhz hardware, partly because of this salami tactics developer attitude. The request for unittests also came a year or so after the original pr. It's a shame that I wasted so much time preparing this contribution.

CurlyMoo:
I reread the whole discussion again, and some requests where just never dealt with. On the 8 of March, the discussion stopped where i requested a manual page, but that was never made, due to license discussions. I don't track PR's to see if developers already implemented all requested parts in a PR. That's why it was never merged.

And as long as a PR isn't updated as requested, the project continious development. That can mean, new demands and guideliness as set. In this case, this means, the unittesting infrastructure and documentation in RST has been implemented while i was waiting for that documentation. Also, the license for new code has changed from GPLv3 to MPLv2.

What you call `salami tactics developer attitude` is something i experience a lot in other open source projects. I just wouldn't label it that way.
- Developers don't track PR's, but you ping them, which saves a lot of time.
- If not all requests are dealt with, they don't bother, which saves a lot of time.
- Projects evolve over time, which requires quite a bit of rewriting of PR code.

janLo:
For the salami tactics: I mean the way you came up with ever new change requests to the original PR, I fixed them - reported back - silence. Then a few months later I asked again and there came new requests that somehow where not present in the original overview. I fixed - reported back - silence. Few months later, same game.

CurlyMoo:
That only really happened once. Between januari 2017 and march 2017. The obvious request and that of @NiekD for documentation was never implemented.

And as i said before, i'm not actively tracking PR's. And in my experience a lot of development don't. At least in the project i contributed to earlier.

PR #330, opened 2017-02-06
Improved status checking in mail.c in accordance to RFC requirements

wo-rasp:
Terminate sendmail actions whenever Error 4xx and 5xx are reported back.

### DIFF
--- a/libs/pilight/core/mail.c
+++ b/libs/pilight/core/mail.c
@@ -209,6 +209,16 @@ starttls:
 			error = -1;
 			goto close;
 		}
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: connection protocol issues - abort");
+			error = -1;
+			goto close;
+		}
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: negative response from remote server - abort");
+			error = -1;
+			goto close;
+		}
 	}
 
 	len = strlen("EHLO ")+strlen(USERAGENT)+3;
@@ -344,18 +354,13 @@ starttls:
 		if(strncmp(recvBuff, "235", 3) == 0) {
 			break;
 		}
-		if(strncmp(recvBuff, "451", 3) == 0) {
-			logprintf(LOG_NOTICE, "SMTP: protocol violation while authenticating");
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: connection protocol issues - abort");
 			error = -1;
 			goto close;
 		}
-		if(strncmp(recvBuff, "501", 3) == 0) {
-			logprintf(LOG_NOTICE, "SMTP: cannot decode response");
-			error = -1;
-			goto close;
-		}
-		if(strncmp(recvBuff, "535", 3) == 0) {
-			logprintf(LOG_NOTICE, "SMTP: authentication failed: wrong user/password");
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: negative response from remote server - abort");
 			error = -1;
 			goto close;
 		}
@@ -385,6 +390,16 @@ starttls:
 		if(strncmp(recvBuff, "250", 3) == 0) {
 			break;
 		}
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: connection protocol issues - abort");
+			error = -1;
+			goto close;
+		}
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: negative response from remote server - abort");
+			error = -1;
+			goto close;
+		}
 		memset(recvBuff, '\0', sizeof(recvBuff));
 	}
 
@@ -411,6 +426,16 @@ starttls:
 		if(strncmp(recvBuff, "250", 3) == 0) {
 			break;
 		}
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: connection protocol issues - abort");
+			error = -1;
+			goto close;
+		}
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: negative response from remote server - abort");
+			error = -1;
+			goto close;
+		}
 		memset(recvBuff, '\0', sizeof(recvBuff));
 	}
 
@@ -436,6 +461,16 @@ starttls:
 		}
 		if(strncmp(recvBuff, "354", 3) == 0) {
 			break;
+		}
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: connection protocol issues - abort");
+			error = -1;
+			goto close;
+		}
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: negative response from remote server - abort");
+			error = -1;
+			goto close;
 		}
 		memset(recvBuff, '\0', sizeof(recvBuff));
 	}
@@ -472,6 +507,16 @@ starttls:
 		if(strncmp(recvBuff, "250", 3) == 0) {
 			break;
 		}
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: connection protocol issues - abort");
+			error = -1;
+			goto close;
+		}
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: negative response from remote server - abort");
+			error = -1;
+			goto close;
+		}
 		memset(recvBuff, '\0', sizeof(recvBuff));
 	}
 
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Again, i want this merged into rewrite.

wo-rasp:
Understood and i will submit 2nd PRs for rewrite and for all changes.
I could have changed the target branch from rewrite back to development but not my source branch development to rewrite, thus I decided to clean up all my branches (doing some git magic) and start submitting all changes in a clean and orderly manner again.
I need to do so as I am using a set of various protocols on my own test_dev branch which basically is an enhanced version of your development branch, whereas development is a plain mirror of yours only.


Von meinem iPhone gesendet

libs/pilight/core/mail.c
```diff
@@ -344,7 +344,7 @@ int sendmail(char *host, char *login, char *pass, unsigned short port, struct ma
 		if(strncmp(recvBuff, "235", 3) == 0) {
 			break;
 		}
-		if(strncmp(recvBuff, "451", 3) == 0) {
+		if(strncmp(recvBuff, "451", 1) == 0) {
```
CurlyMoo: Why did you change this to 1?

libs/pilight/core/mail.c
```diff
@@ -359,6 +359,11 @@ int sendmail(char *host, char *login, char *pass, unsigned short port, struct ma
 			error = -1;
 			goto close;
 		}
+		if(strncmp(recvBuff, "500", 1) == 0) {
```
CurlyMoo: Likewise

libs/pilight/core/mail.c
```diff
@@ -359,6 +359,11 @@ int sendmail(char *host, char *login, char *pass, unsigned short port, struct ma
...
+			logprintf(LOG_NOTICE, "SMTP: neg. response - abort");
```
CurlyMoo: Can you be a bit more verbose here?

libs/pilight/core/mail.c
wo-rasp: In order to add all other 4xx reply codes, as all of them are fatal.

libs/pilight/core/mail.c
wo-rasp: All 12 currently known 5xx reply codes are more or less fatal as well

libs/pilight/core/mail.c
wo-rasp: A 5xx reply code is fatal, for some replys (most notable 551), parsing of the reponse is required, but not implemented, so i decided to simply terminate the request.

CurlyMoo:
So shouldn't we implement the responses to non-fatal 4xx and 5xx messages?

libs/pilight/core/mail.c
CurlyMoo: So this is redundant to the previous 5xx check?

wo-rasp:
It is always good to do this kind of code review (in particular comments on my remark under 1c are appreciated)

After reviewing the subject my answer to your is NO due to the following reasons:

 

1.       The original intend for this modification was the monitored behaviour that the pilight service terminates itself due to high CPU load, one of the reasons was the not handled response code 452.

a.       In my opinion there are many more state condition of the current mail subsystem implementation causing it to stay in this particular loop (L340 ff) for more than 10 secs.

                                                               i.      I do know of at least 14 4xx and 17 5xx codes in common use, and as far as I know there may even more be implemented.

b.      I propose to add exceptions first and if they are not applicable, to treat all other 4xx, 5xx conditions as fatal.
(Example: Reply 452 would allow to requeue a message, thus we can use it and implement a queue for that that message …..) 

c.       Based on current experience there most be more unidentified conditions present, as those modifications led to an improvement but did not fully solve the high CPU load scenario in case the mail server is not available.
IMHO there is at least another not properly handled condition present in this while loop (starting at line 340), it is the assumption that once authentication is successful, we always do get reply code 235, resp. that we do get certain 4xx or 5xx replies. I agree that we should get the 235 message, but what happens if we do not ?

                                                               i.      In general,  I do propose to

1.       either add a 50mS delay to this loop and to wait until we get a 235, 4xx, 5xx reply

2.       or to exit with an error condition if we do not get a 235, 4xx, 5xx reply within 60 seconds

3.       or to add a loop counter and to exit this loop if we do not get the 235, 4xx, 5xx reply within 10 replies

                                                             ii.      An additional break criteria should be applied to the following while loops as well:

1.       Starting  at Line 230, 284, 386, 412, 438, 473, and 499 

d.      If the user is unable to diagnose the issue, running pilight daemon in debug mode already reveals the reason codes returned, Line 342 ff. has implemented the required logic.

 

 

Gesendet: Dienstag, 7. Februar 2017 06:47
An: pilight/pilight
Cc: wo-rasp; Author
Betreff: Re: [pilight/pilight] Improved status checking in mail.c in accordance to RFC requirements (#330)

 

So shouldn't we implement the responses to non-fatal 4xx and 5xx messages?

wo-rasp:
Yes and No, I kept the 501 and 535 codes, in order to keep the two response messages in the log unchanged.

 

I have not fully analysed the AUTH phase, leading to the expectation that a 235 REPLY is returned.

In RFC 2554 the following states are discussed:

 

S: 220 smtp.example.com ESMTP server ready
C: EHLO jgm.example.com
S: 250-smtp.example.com
S: 250 AUTH CRAM-MD5 DIGEST-MD5
C: AUTH FOOBAR
S: 504 Unrecognized authentication type.
C: AUTH CRAM-MD5
S: 334
PENCeUxFREJoU0NnbmhNWitOMjNGNndAZWx3b29kLmlubm9zb2Z0LmNvbT4=
C: ZnJlZCA5ZTk1YWVlMDljNDBhZjJiODRhMGMyYjNiYmFlNzg2ZQ==
S: 235 Authentication successful. 

 

As discussed separately and based on this example, while waiting for the 235 REPLY,  we may want to handle the REPLY code 504 as an exemption:

 

if(strncmp(recvBuff, "235", 3) == 0) {

    break;

}         

if(strncmp(recvBuff, "504", 3) == 0) {

    logprintf(LOG_NOTICE, "SMTP: %s”, recvBuff);

    break;

}

if(strncmp(recvBuff, "500", 1) == 0) {

    logprintf(LOG_NOTICE, "SMTP: %s”, recvBuff);

    error = -1;

    goto close;

} 

 

But looking at the mail.c code, we are not re-entering the authentication phase anyway, and based on not getting the expected response (“MAIL from”) we can treat that reply here as fatal already and go to close.

 

Question:
Does logprintf() run in task context (e.q. is it fully executed before we reach the last statement at the end of the while loop:

memset(recvBuff, '\0', sizeof(recvBuff));

?



If not we may want to keep the static text messages. 

The abbreviated revised code would then look like

 

    if(strncmp(recvBuff, "235", 3) == 0) {

        break;

    }

    if(strncmp(recvBuff, "500", 1) == 0) {

        logprintf(LOG_NOTICE, "SMTP: neg. response - abort");

        error = -1;

        goto close;

    }

//  … add a while loop exit criteria as discussed separately …

    memset(recvBuff, '\0', sizeof(recvBuff));

}

 

Gesendet: Dienstag, 7. Februar 2017 06:48
An: pilight/pilight
Cc: wo-rasp; Author
Betreff: Re: [pilight/pilight] Improved status checking in mail.c in accordance to RFC requirements (#330)

 

@CurlyMoo commented on this pull request.

  _____  

In libs/pilight/core/mail.c <https://github.com/pilight/pilight/pull/330> :

> @@ -359,6 +359,11 @@ int sendmail(char *host, char *login, char *pass, unsigned short port, struct ma
                       error = -1;
                       goto close;
               }
+              if(strncmp(recvBuff, "500", 1) == 0) {
+                      logprintf(LOG_NOTICE, "SMTP: neg. response - abort");

So this is redundant to the previous 5xx check?

CurlyMoo:
I have to say that responding by email makes the comments pretty unreadable :smile: Can you please reformat the whole thing so it is easier for me to respond.

wo-rasp:
Updated and added handling of 4xx, 5xx reply codes to all reply checking loops.

CurlyMoo:
I'd rather see a `4xx` and a `5xx` in the code. That makes it more clear we are checking for all 400 and 500 range codes. Also, can you commit under your account.

wo-rasp:
Account issue fixed.
The development branch and the rewrite branch are not compatible, a specially adapted version for the rewrite branch is submitted.

4xx, 5xx:
Why not just 4 and 5 ?

CurlyMoo:
> 4xx, 5xx:
> Why not just 4 and 5 ?

Because 4xx tells me in a glance that we are covering a whole range of codes. In case of 4 i have to figure that out.

wo-rasp:
Replaced 400 and 500 string comparison with 4xx and 5xx to highlight that only the 1st byte is part of the comparison to identify all response codes between 400 and 599 as being unexpected and not recoverable, despite the fact that for some response codes a later recovery of the server is not impossible, but as that would mean to do housekeeping for all those mails we discard them gracefully with an error message, being different for 4xx codes and 5xx codes.

libs/pilight/core/mail.c
```diff
@@ -209,6 +209,16 @@ int sendmail(char *host, char *login, char *pass, unsigned short port, struct ma
 			error = -1;
 			goto close;
 		}
+		if(strncmp(recvBuff, "4xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: protocol issues - abort");
+			error = -1;
+			goto close;
+		}
+		if(strncmp(recvBuff, "5xx", 1) == 0) {
+			logprintf(LOG_NOTICE, "SMTP: neg. response - abort");
```
CurlyMoo: Can you just write this fully instead of using a abbreviation?

wo-rasp:
Marked 4xx as "conenction protocol issues" and 5xx errors as "negative resonse from remote server"

PR #332, opened 2017-02-10
Improved status checking in mail.c in accordance to RFC requirements

wo-rasp:
Added Error Checking for 4xx and 5xx codes to mail.c rewrite branch


CurlyMoo:
I'd rather see a `4xx` and a `5xx` in the code. That makes it more clear we are checking for all 400 and 500 range codes. Also, can you commit under your account.

I also miss the unit test to check if these errors are correctly handled.

wo-rasp:
The function call tells specifically, that only the 1 digit is checked. 400 is as good as 4xx or 4

Unittest:
Can you elaborate how that unit test shall be look like ?
Which mail server implementation and which RFC paper from the various smtp implemantations around shall serve as a base ?

CurlyMoo:
> Unittest
What i did was using telnet to talk to actual mailservers and simulating it in the unit test. It did not use any protocol papers.

So, simulate a valid 4xx and a valid 5xx response in the unit test and check if the mail function indeed fails as expected. I believe the 4xx is already implemented.

libs/pilight/core/mail.c
```diff
@@ -206,7 +206,7 @@ static void read_cb(uv_poll_t *req, ssize_t *nread, char *buf) {
 					}
 				}
 
-				if(val == 501 && ch == 32) {
+				if( (val >= 400 && val <= 599) && ch == 32) {
```
pilino1234: Spurious space before brackets, otherwise :+1:

wo-rasp: done

CurlyMoo:
Any news on the unit test?

pilino1234:
bump, and needs rebasing

CurlyMoo:
And it needs additional tests.

PR #334, opened 2017-02-24
Respect node UUID configuration also for RF433 and RF868 devices.

nullcheck:
Small change to allow pilight's adhoc network feature to respect optionally configured node UUIDs also for RF433 and RF868 devices. This can be used to prevent interference if multiple pilight nodes that are relatively close to one another try to send the same command at the same time.
The same change / enhancement was also announced for the pilight rewrite.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -1083,7 +1083,9 @@ static int control_device(struct devices_t *dev, char *state, struct JsonNode *v
 	/* Construct the right json object */
 	json_append_member(code, "protocol", jprotocols);
 	if(dev->dev_uuid != NULL && (dev->protocols->listener->hwtype == SENSOR
-	   || dev->protocols->listener->hwtype == HWRELAY)) {
+	   || dev->protocols->listener->hwtype == HWRELAY
+	   || dev->protocols->listener->hwtype == RF433
+	   || dev->protocols->listener->hwtype == RF868)) {
 		json_append_member(code, "uuid", json_mkstring(dev->dev_uuid));
 	}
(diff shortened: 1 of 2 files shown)


CurlyMoo:
Did you copy this from rewrite branch? I already implemented this feature in there.

nullcheck:
No, I got the info that you want to change this for your rewrite from a forum post you made some time ago about the rewrite's features. I actually did not think about looking into the rewrite's code since I figured it would be too different to compare.
I made this change back in December when I added another raspi to extend the range of my pilight setup. Since I had quite bad interference when controlling some of my RF433 plugs, I took a look at the code and was happy to find that the required change was minimal. My plan was to test it a bit longer before submitting the PR, but then I forgot about it entirely until recently. 
So I can say that I have been testing the change on a daily basis now for about two months and did not face any issues. That is for RF433 only though, I don't own any RF868 devices.

CurlyMoo:
I love the effort you took to backport this already, but i still think it would be good if you would compare your changes with the ones i did in the rewrite branch.

madattack:
Hello,

World it bereits possible to include this in a release in the bear future? Would love to have this feature . 

Cheers

CurlyMoo:
> Would it be possible to include this in a release in the bear future? Would love to have this feature .

Already included in rewrite.

pilino1234:
This feature is included both in development and stable, might be possible to close this PR now.

CurlyMoo:
Yes. Thanks for the PR!

PR #339, opened 2017-03-16
Livolo protocol (send only)

leppa:
Here's an implementation of the protocol used by Livolo wireless switches (and their clones).

Protocol description was taken from here: http://platenspeler.github.io/DeveloperGuide/433Transmitters/433messaging.html

Remote keycodes were taken from here: https://github.com/platenspeler/LamPI-2.0/tree/master/transmitters/livolo

Unfortunately, it's impossible to implement reading of the protocol with current pilight architecture, because livolo protocol doesn't have footer (which is used by pilight to extract a pulsetrain). It means that only send functionality is supported ATM. I'm open to suggestions on how to implement parsing of the protocol.

The protocol was tested with several switches that I have at home and works well with all of them.

### DIFF
--- a/CMakeConfig.txt
+++ b/CMakeConfig.txt
@@ -53,6 +53,7 @@ set(PROTOCOL_KAKU_SCREEN ON CACHE BOOL "support for the new KlikAanKlikUit scree
 set(PROTOCOL_LM75 ON CACHE BOOL "support for the lm75 i2c temperature sensor")
 set(PROTOCOL_LM76 ON CACHE BOOL "support for the lm76 i2c temperature sensor")
 set(PROTOCOL_LIRC ON CACHE BOOL "support for the Lirc API")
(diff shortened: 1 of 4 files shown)


CurlyMoo:
What a great clean PR!

Some comments (for now):
1. Can you also create a wiki page for this device.
2. Can you also do a PR for the rewrite branch including the unit test.

leppa:
> 1. Can you also create a wiki page for this device.

Just created it: https://wiki.pilight.org/doku.php/livolo_switch_v7_0

I'm using pilight together with Home Assistant, and don't have any switches in the pilight config. So I didn't create config section. I'll test the whether config works when I have time, and update the wiki.

> 2. Can you also do a PR for the rewrite branch including the unit test.

Are there any big changes needed in the rewrite branch?
This might take a bit, because I've been short on spare time recently.

CurlyMoo:
> Are there any big changes needed in the rewrite branch?

In regard to protocol modules not.

PR #340, opened 2017-03-21
Not all fields of protocol_t were initialized in protocol_register()

leppa:
Also fixes build error when WEBSERVER_HTTPS is not defined.

### DIFF
--- a/tests/protocols_433.c
+++ b/tests/protocols_433.c
@@ -88,7 +88,9 @@ static void test_protocols_433(CuTest *tc) {
 			/*
 			 * Check if the validator worked
 			 */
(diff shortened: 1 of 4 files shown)


tests/protocols_433.c
```diff
@@ -88,7 +88,9 @@ static void test_protocols_433(CuTest *tc) {
 			/*
 			 * Check if the validator worked
 			 */
-			CuAssertIntEquals(tc, tests[x].raw[i].validate, protocol->validate());
+			if(protocol->validate) {
```
CurlyMoo: Can you add a `!= NULL`?

leppa: Done

PR #341, opened 2017-03-21
Livolo protocol (sending only) [rewrite branch]

leppa:
As requested in #339, here's my change on top of the `rewrite` branch with two basic tests.


CurlyMoo:
The tests will be used in the future to make sure protocols keep working in case i will change the protocol code. Are you confident you've added enough tests?

leppa:
There's not much to test, actually, because it's send only and the protocol is simple.

CurlyMoo:
One thing i forgot. Can we update the license to MPLv2 in both development and rewrite?

leppa:
No problem. Feel free to update it.

CurlyMoo:
Can you maybe do a new PR with the new license?

PR #348, opened 2017-04-10
added ds_4 protocol

torwag:
First draft of a protocol for magnetic contact sensors obtained from [China](https://www.aliexpress.com/store/product/5pcs-433MHz-EV1527-two-way-wireless-intelligent-door-window-sensor-APP-control-wifi-door-detector-for/725200_32642367411.html).  
The come with a Chinese uC STC15W100, which might help to identify clones of this unit with the same protocol.
The unit claims two-way communication however, that actually means it sends out both events, opening and closing, as well as a tempering event. It also claims low battery but I could not yet figure that out. Each event is send out 15-20 times (did not count). 

The protocol is very similar to the EV1527, yet different enough that I could not add it easily without messing the EV1527 code. 
Hence a new protocol. From Raw mode inspection and testing
```
AW RF signal (50): c:01100101100110010101100101010110101010100110011012;p:1395,446,13999@
 
Sensor 1:
01100101010110011010010110101001101010100101011012 on
01100101010110011010010110101001101010100110011012 off
01100101010110011010010110101001101010101001010112 tamper on & off
01100101010110011010010110101001101010101001010112 tamper on & on
01100101010110011010010110101001101010100101011012 tamper off & on

Sensor2:
01100101100110101001011010100101101010100101011012 on
01100101100110101001011010100101101010100110011012 off

Sensor3:
01100101100110010101100101010110101010100110011012 off

removed the last two pulses (footer) and decode with 10 = 0 and 01 = 1 
results in the following bit codes

10111101 00110001 0000 1110 on sensor 1
10111101 00110001 0000 1010 off sensor 1
10111101 00110001 0000 0111 tamper on & off sensor 1
10111101 00110001 0000 0111 tamper off & on sensor 1
10111101 00110001 0000 1110 tamper off & on
10110100 01100011 0000 1110 on sensor 2
10110100 01100011 0000 1010 off sensor 2
10110101 11011110 0000 1010 off sensor 3

First 16 bit seem to be the unitcode address:
then 4 unused bit (battery?)
followed by 4 data bits in the form tamper on/off 1 tamper 
```
There seems to be no combination, each event sends only the event code without taking care of the state of the other bits. At least that allows for easy binary number checking to identify the event.
Seller claims low battery function, but I could not yet discover this in the datagram.
The PCB of the unit is named DS-4, thus I named the protocol like this. Seller claim it is a sensor working for G90B G90E  alarm systems. As a slightly more known brand Kerui is selling similar sensors. However, I could not yet test the protocol as I do not own a Kerui contact sensor. Changes are high that they are compatible. 

Compilation of pilight is okay, but I could not test it since I  have no hardware. I used a version of the [MQTT433Gateway](https://github.com/puuu/MQTT433gateway), which in turn uses a port of pilight for the esp8266. I tested the patch on the ESP8266 with MQTT433gateway and all events were shown correctly.
Before, there was no event at all, thus I believe the protocol was missing. I was told to submit a protocol patch upstream, to make sure digital cookies are send to those who earned them ;)

This is my first contribution to pilight and hence maybe not in the correct formal order. I am happy to improve the patch or add changes as required.

### DIFF
--- a/CMakeConfig.txt
+++ b/CMakeConfig.txt
@@ -22,6 +22,7 @@ set(PROTOCOL_DATETIME ON CACHE BOOL "support for the (NTP) date and time protoco
 set(PROTOCOL_DAYCOM ON CACHE BOOL "support Daycom switch (Pollin)")
 set(PROTOCOL_DS18B20 ON CACHE BOOL "support for the 1-Wire ds18b20 temperature sensor")
 set(PROTOCOL_DS18S20 ON CACHE BOOL "support for the 1-Wire ds18s20 temperature sensor")
+set(PROTOCOL_DS_4 ON CACHE BOOL "support for the DS-4 contact sensor (deviation from EV1527")
 set(PROTOCOL_DHT11 ON CACHE BOOL "support for the DHT11 temperature-humidity sensor")
 set(PROTOCOL_DHT22 ON CACHE BOOL "support for the DHT22/AM2302 temperature-humidity sensor")
 set(PROTOCOL_DIO_SWITCH ON CACHE BOOL "support for the D-IO switch protocol")
--- a/libs/pilight/protocols/433.92/ds_4.c
+++ b/libs/pilight/protocols/433.92/ds_4.c
@@ -1,0 +1,120 @@
+/*
+	Copyright (C) 2017 Torwag
+	
+	This file is part of pilight.
+
+	pilight is free software: you can redistribute it and/or modify it under the
+	terms of the GNU General Public License as published by the Free Software
+	Foundation, either version 3 of the License, or (at your option) any later
+	version.
+
+	pilight is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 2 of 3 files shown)


CurlyMoo:
Master PR's are not accepted. Please read the contribution guide about PR rules.

I would also like you to discuss the protocol on the pilight forum and please do test this thing in real life.

Lastly, can you write the wiki page and also do the PR for the rewrite branch.

torwag:
Thanks for the infos. I will check about the contribution guideline. As I said this PR was just a first sketch.
As for testing. I can say it works together with the MQTT433Gateway, which in turn uses the pilight protocols. If further testing is needed, I need to set-up a pilight system first. Alternatively, I might be able to find someone within driving distance to test the patch.

I started a thread in the [forum](https://forum.pilight.org/Thread-DS-4-protocol-magnetic-switch-EV1527-deviation). Wiki and other PRs might follow according to the results of the thread.

Thanks again

PR #359, opened 2017-06-24
repair broken i2c devices / new i2c-path config entry

GreenMarc:
Hi,

has anyone tested the i2c devices (bmp180,lm75,lm76) after the wiringx update ? It dont works because the new i2c-path config-entry-code has a typo..... see attached patch...

Greetz
Marc 

PS: this typo is in the rewrite code also...


pilino1234:
Please submit a second PR to the rewrite branch in that case, then they can both be merged together 😄 

I haven't tested this because I don't own any I2C pilight hardware, but the patch LGTM.

:+1:

PR #360, opened 2017-06-24
using localtime instead of UTC in the logfile

GreenMarc:
Hi,
in the logfiles pilight.err & pilight.log the timestamp is 2 hours ahead of my systemtime. The datetime protocol dont have this issue...
My raspbian time is set to localtime and in-sync with ntp....

The attached patch changed the timeformat in the logfile to the localtime and is now the same like the systemtime...

Greetz
Marc

### DIFF
--- a/libs/pilight/core/log.c
+++ b/libs/pilight/core/log.c
@@ -169,10 +169,10 @@ void logprintf(int prio, const char *format_str, ...) {
 		gettimeofday(&tv, NULL);
 #ifdef _WIN32
 		struct tm *tm1;
-		if((tm1 = gmtime(&tv.tv_sec)) != 0) {
+		if((tm1 = localtime(&tv.tv_sec)) != 0) {
 			memcpy(&tm, tm1, sizeof(struct tm));
 #else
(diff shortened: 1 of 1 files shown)


pilino1234:
IIRC, this is intentional in order to avoid clusterfucks in case of an NTP update:

https://www.pilight.org/post/pilight-70 (look in "Bugfixes")

also 9541d200d372141e447de269c4d010fb248dd867

GreenMarc:
ROFL....ok... i dont found this...

Do you have more infos of this TZ shifts ? Any forumposting etc ?

pilino1234:
I couldn't find anything with forum search (which, admittedly, is notoriously bad at finding stuff), but google doesn't help either. Afaik, there is no thread about this topic, only single semi-offtopic postings in other threads that are basically impossible to find.

Tbh, I'm not familiar with the reason behind this decision, but I'm fairly sure it's a good one. @CurlyMoo probably knows though.

CurlyMoo:
Indeed, devices similar to the RPi are not very reliable in case of system time, because they don't have an onboard RTC. That's why we store logs in UTC.

PR #361, opened 2017-06-24
repair broken i2c devices / new i2c-path config entry

GreenMarc:
here is the rewrite patch


pilino1234:
Thanks, now we just have to wait for @CurlyMoo to merge these

CurlyMoo:
Can you check the unittest for these devices. These do work, so we should test for faulty settings. Which means extending the PR with more extensive testing.

GreenMarc:
Sorry, i dont have any of theese devices. 

I updated the code of my own i2c devices (BH1750, TSL2561 & BME280) against the current development branch. I orientated myself at your code of the bmp180/lm75/lm76 and it wont work. So im searching in your code whats wrong... I found the typo in the devel branch and found it also in the rewrite branch...

For your tests i must rewrite the code for my i2c devices....

CurlyMoo:
> Sorry, i dont have any of theese devices.

The good thing is that the unit tests are all done without the actual devices but with the `i2c-stub` module. Just check the `protocol-i2c.c` unit test. Secondly, we are just fixing a regex here.

PR #362, opened 2017-06-30
Fixed ntp bugs

NiekD:
1. Check for failing socket send- and receive was wrong. If receive failed, the result was processed anyway, causing "random" diff values and, as a result, weird datetime values. Fixed by changing the check for the return values.

2. Check for msg.refid was wrong, because refid can have a negative value. In such cases, the result of the ntp request was treated as invalid, causing the "could not sync with ntp server..." message. Fixed by changing the check for the msg.refid value.

### DIFF
--- a/libs/pilight/core/ntp.c
+++ b/libs/pilight/core/ntp.c
@@ -182,13 +182,13 @@ void *ntpthread(void *param) {
 					setsockopt(sockfd, SOL_SOCKET, SO_RCVTIMEO, (char *)&tv, sizeof(struct timeval));
 #endif
 					msg.li_vn_mode=227;
-					if(sendto(sockfd, (char *)&msg, 48, 0, (struct sockaddr *)&servaddr, sizeof(servaddr)) < -1) {
+					if(sendto(sockfd, (char *)&msg, 48, 0, (struct sockaddr *)&servaddr, sizeof(servaddr)) < 0) {
 						logprintf(LOG_NOTICE, "error in ntp sending");
 					} else {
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Have you checked the rewrite implementation as well?

NiekD:
I did not check the rewrite implementation, but I can try if you wish.

I discovered the fact that the refid can be negative just by debugging the ntp.c code, but I think it is implicitly confirmed in this document:  http://support.ntp.org/bin/view/Support/RefidFormat

you can see that the refid in fact is not really defined as a 32 bit integer, but just as 4 bytes representing the time source. That can either be:

- the binary representation of a 4 character ascii string,
- or an ip4 address, 
- or the four first bytes of the md5 hash of an ip6 address.

If the value the high bit of the first byte of an md5 has happens to be 1, the resulting refid will be negative.

I guess that that explains the fact that the datetime issue has started rather recently and has increased over time: time servers migrating from ip4 to ip6.

CurlyMoo:
Yes, please check rewrite as well. I will make sure it compiles again :smile:

NiekD:
I had a look at the rewrite code and the refid fix ( changing msg.refid > 0 to msg.refid != 0) can simply be applied to the ntp.c rewrite (at line 143) too. Since you redesigned the send and receive code completely, my  fix for failing socket send and -receive is irrelevant for the rewrite. 

I could create a PR for the rewrite, but at this moment I have no system available to test the fix in the rewrite myself. Although it is just a small modification, I think it is wrong to issue a PR without having tested.

So maybe in this case it is better if you (or someone else who can test it) applies the fix. Otherwise it will unfortunately take quite some time before I will be able to do that.

PR #345, opened 2017-03-24
Create manpages for all pilight programs

pilino1234:
Quite self-explanatory

Now automatically install manpages to `/usr/share/man/man1`. Tested and confirmed working, but comments on the CMake stuff are welcome. 

Use `sudo make install` to test. Output should look like this:
```-- Installing: /usr/share/man/man1/pilight-sha256.1
-- Installing: /usr/share/man/man1/pilight-control.1
-- Installing: /usr/share/man/man1/pilight-raw.1
-- Installing: /usr/share/man/man1/pilight-debug.1
-- Installing: /usr/share/man/man1/pilight-flash.1
-- Installing: /usr/share/man/man1/pilight-send.1
-- Installing: /usr/share/man/man1/pilight-uuid.1
-- Installing: /usr/share/man/man1/pilight-receive.1
-- Installing: /usr/share/man/man1/pilight-daemon.1
```
Then use `man` to view the manpages.

Will squash soon.

___
~~(These are currently not installed though, I'm still tinkering with CMakeLists.txt to make it do what I want it to. If anyone knows how to make CMake install manpages, I would appreciate any help!)~~

PR #363, opened 2017-07-08
Split --nodaemon option into --nodaemon and --verbose

schaal:
Don't change the log level when using --nodaemon. Add --verbose option to set the log level to debug.

This is for better integration with systemd to be able to prevent pilight from forking without increasing the log level.

### DIFF
--- a/daemon.c
+++ b/daemon.c
@@ -195,7 +195,7 @@ static char *master_server = NULL;
 static unsigned short master_port = 0;
 
 static char *configtmp = NULL;
-static int verbosity = LOG_DEBUG;
+static int verbosity = LOG_INFO;
 struct socket_callback_t socket_callback;
 
 #ifdef _WIN32
@@ -2082,7 +2082,7 @@ int start_pilight(int argc, char **argv) {
 	struct ssdp_list_t *ssdp_list = NULL;
 
 	char buffer[BUFFER_SIZE];
-	int itmp = 0, show_default = 0, show_version = 0, show_help = 0;
+	int verbosity_changed = 0, show_default = 0, show_version = 0, show_help = 0;
 #ifndef _WIN32
 	int f = 0;
 #endif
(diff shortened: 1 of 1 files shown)


CurlyMoo:
All pilight core options are capatilized, so can you capatalize the `V` as well?

schaal:
> All pilight core options are capatilized, so can you capatalize the `V` as well?

Turns out `V` is already taken by the version option. I choose `I` instead for the short option as in "debug **I**nformation" (as `D` is already taken as well).

CurlyMoo:
What about just leaving the `I` out and keep the debuglevel option.

CurlyMoo:
Maybe we should add a `--debug` option so the user will can have a `nodaemon` mode and `debuglevel` at once?

schaal:
This now does the following:

* Don't set the log level to debug when nodaemon==1
* Don't override the log level set in the config file when nodaemon==1
* If any command line option is given that changes the verbosity (stacktracer, debuglevel, etc...) ignore the value in the config file.

schaal:
> Maybe we should add a --debug option so the user will can have a nodaemon mode and debuglevel at once?

My preference would be to just have the two options independent of each other, but I can add the debug option if you want.

CurlyMoo:
> My preference would be to just have the two options independent of each other, but I can add the debug option if you want.

Let's see what others think.

pilino1234:
I quite like the current state of this PR as described here: https://github.com/pilight/pilight/pull/363#issuecomment-313909217

Although I have a couple of improvments:

- I don't like the name of the `--debuglevel` option currently, I think `verbosity` would be better.
- The possible debuglevels should be listed somewhere, maybe in the usage summary since `man` pages don't exist ([yet](https://github.com/pilight/pilight/pull/345)). A different alternative would be to use chaining options to change verbosity, so that `-vvvvv` would be more verbose than `-vv`, for example.
- Short options should be added for any of these that don't have them yet, because combining `--nodaemon`/`-D` with what `--debuglevel` does now is done quite commonly and should be easily accessible. Otherwise, adding a third option like what @CurlyMoo suggests (https://github.com/pilight/pilight/pull/363#issuecomment-313911447) should be done.

CurlyMoo:
> I don't like the name of the --debuglevel option currently, I think verbosity would be better.

The `--debuglevel` is currently an option for developers which give information about the internals of pilight, which is much more verbose then the regular `LOG_*` verbosities.

> A different alternative would be to use chaining options to change verbosity, so that -vvvvv would be more verbose than -vv, for example.

This way of having stacked shortcodes are not supported (yet).

pilino1234:
> The --debuglevel is currently an option for developers which give information about the internals of pilight, which is much more verbose then the regular LOG_* verbosities.

Whoops, that was a misread on my part. Basically, what I'm wondering is how do I achieve the same behaviour as what `pilight-daemon -D` did previously?

CurlyMoo:
> Whoops, that was a misread on my part. Basically, what I'm wondering is how do I achieve the same behaviour as what pilight-daemon -D did previously?

That would require a `--verbosity` option set to `LOG_DEBUG` and the `-D` option combined.

pilino1234:
That's what I'd like to have, yes.

schaal:
I incorporated some of the feedback and it now looks like this:

* Keep the current behavior of the `-D` option (no forking and set log level to debug), but rename the long option to `--debug`.
* Add option `--nodaemon`/`-N` which prevents forking, but doesn't change the log level
* Command line options override the log level set in the config file

CurlyMoo:
Maybe we can change `--nodaemon` to `--foreground` and `-F`. That's how it's called in other programs.

schaal:
> Maybe we can change --nodaemon to --foreground and -F. That's how it's called in other programs.

Done

pilino1234:
:+1:

daemon.c
```diff
@@ -2329,6 +2336,19 @@ int start_pilight(int argc, char **argv) {
 
 	registerVersion();
 
+	// let verbosity level from command line take precedence over config file
+	if(!verbosity_changed) {
```
CurlyMoo: Can you make this `if(verbosity_changed == 0) {`

CurlyMoo:
Can you squash everything into one commit?

CurlyMoo:
Thanks!

PR #367, opened 2017-07-22
Port pilight-receive filtering feature to rewrite

pilino1234:
Also removes tabs at the end of some lines

I'm thinking about adding a second type of filtering, allowing users to filter **for** a specific protocol(s). This is useful with weather sensors, when the ID has changed after changing the batteries, so that users don't have to list every protocol that isn't wanted in the output of pilight-receive, but rather specify which one(s) they _want_ to see.
i think I would do it like this:

 * Repurpose the `-F` flag to become a "real" filter, i.e. it only outputs the specified protocols (`-F alecto_wsd17` would only output `alecto_wsd17` signals). Listing several protocols should be supported I think - it doesn't add a lot of bloat but it might be useful in certain situations.
 * Add a `-X` (`--eXclude`) flag to list protocols that should be excluded (like what the -F flag does currently). `-X datetime,cpu_temp` would not output `datetime` or `cpu_temp` messages.

However, I'm uncertain about the repurposing of the -F flag to become the exact opposite of the current behaviour. Thoughts?

### DIFF
--- a/receive.c
+++ b/receive.c
@@ -28,6 +28,9 @@
 #include "libs/pilight/core/ssdp.h"
 #include "libs/pilight/protocols/protocol.h"
 
+#define INCLUDE		1
+#define EXCLUDE		2
+
 static uv_tty_t *tty_req = NULL;
 static uv_signal_t *signal_req = NULL;
 
@@ -40,7 +43,7 @@ static unsigned short stats = 0;
 static unsigned short connecting = 0;
 static unsigned short connected = 0;
 static unsigned short found = 0;
-static unsigned short filteropt = 0;
+static unsigned short filtertype = 0;
 
 typedef struct ssdp_list_t {
 	char server[INET_ADDRSTRLEN+1];
@@ -99,6 +102,40 @@ static void alloc_cb(uv_handle_t *handle, size_t len, uv_buf_t *buf) {
 	memset(buf->base, 0, len);
 }
 
+static void output(struct JsonNode *json) {
+	if(filtertype != 0) {
+		int filtered = 0, j = 0;
+		char *protocol = NULL;
+		if(filtertype == INCLUDE) {
+			filtered = 1;
+		}
+		json_find_string(json, "protocol", &protocol);
+		for(j=0;j<nrfilter;j++) {
+			if(strcmp(filters[j], protocol) == 0) {
+				if(filtertype == INCLUDE) {
+					filtered = 0;
+					break;
+				} else if(filtertype == EXCLUDE) {
+					filtered = 1;
+					break;
+				}
+			}
+		}
+		if(filtered == 0) {
+			char *out = json_stringify(json, "\t");
+			printf("%s\n", out);
+			json_free(out);
+		}
+	} else {
+		char *out = json_stringify(json, "\t");
+		printf("%s\n", out);
+		json_free(out);
+	}
+	json_delete(json);
+
+	return;
+}
+
 static void on_read(int fd, char *buf, ssize_t len, char **buf1, ssize_t *len1) {
 	if(strcmp(buf, "1") != 0 &&
 	   strncmp(buf, "{\"status\":\"success\"}", 20) != 0) {
@@ -109,10 +146,7 @@ static void on_read(int fd, char *buf, ssize_t len, char **buf1, ssize_t *len1) 
 				for(i=0;i<n;i++) {
 					struct JsonNode *json = json_decode(array[i]);
 					if(json != NULL) {
-						char *out = json_stringify(json, "\t");
-						printf("%s\n", out);
-						json_delete(json);
-						json_free(out);
+						output(json);
 					} else {
 						logprintf(LOG_ERR, "invalid JSON received: %s", buf);
 					}
@@ -121,10 +155,7 @@ static void on_read(int fd, char *buf, ssize_t len, char **buf1, ssize_t *len1) 
 			} else {
 				struct JsonNode *json = json_decode(*buf1);
 				if(json != NULL) {
-					char *out = json_stringify(json, "\t");
-					printf("%s\n", out);
-					json_delete(json);
-					json_free(out);
+					output(json);
 				} else {
 					logprintf(LOG_ERR, "invalid JSON received: %s", buf);
 				}
@@ -177,7 +208,7 @@ static void connect_to_server(char *server, int port) {
 	}
 
 	uv_timer_init(uv_default_loop(), socket_timeout_req);
-	uv_timer_start(socket_timeout_req, timeout_cb, 1000, 0);	
+	uv_timer_start(socket_timeout_req, timeout_cb, 1000, 0);
 }
 
 static int select_server(int server) {
@@ -282,13 +313,13 @@ static void main_loop(int onclose) {
 	if(onclose == 1) {
 		signal_cb(NULL, SIGINT);
 	}
-	uv_run(uv_default_loop(), UV_RUN_DEFAULT);	
+	uv_run(uv_default_loop(), UV_RUN_DEFAULT);
 	uv_walk(uv_default_loop(), walk_cb, NULL);
 	uv_run(uv_default_loop(), UV_RUN_ONCE);
 
 	if(onclose == 1) {
 		while(uv_loop_close(uv_default_loop()) == UV_EBUSY) {
-			uv_run(uv_default_loop(), UV_RUN_DEFAULT);	
+			uv_run(uv_default_loop(), UV_RUN_DEFAULT);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


pilino1234:
The coveralls command seems to fail (from travis log):
```
$ coveralls-lcov --repo-token ${COVERALLS_TOKEN} coverage-filtered.info
Too many arguments! <>
  Usage: coveralls-lcov [options] coverage.info
  e.g. coveralls-lcov -v coverage.info
    -t, --repo-token=TOKEN           Repository token
    -s, --service-name=SERVICENAME   Service name
        --service-job-id=JOB_ID      Service job id. ex. TRAVIS_JOB_ID
        --retry=N                    Retry to POST N times (default: 3)
        --delay=N                    Delay in N secs when retry (default: 3)
        --source-encoding=ENCODING   Source files encoding  (default: UTF-8)
    -v, --verbose                    Print payload
    -n, --dry-run                    Dry run
```

CurlyMoo:
Agreed on the -X and -F

> The coveralls command seems to fail (from travis log):

Don't see any failure?

pilino1234:
Ok, will update the PR

> Don't see any failure?

Yeah, nvm. I didn't realize the last coveralls command is for pushing the coverage status to coveralls service. 
Might be good to add a check for PR builds using the `TRAVIS_EVENT_TYPE` environment variable, see https://docs.travis-ci.com/user/environment-variables/. It if it's value is `pull_request`, don't run the last coveralls command to push coverage status.

pilino1234:
Updated! Has a lot of code duplication imo, but I couldn't figure out a way to factor out more code. Maybe I'm just too tired though 😉

receive.c
```diff
@@ -378,12 +442,24 @@ int main(int argc, char **argv) {
...
 			break;
+			case 'X':
+				if(filteropt == 0) {
+					if((excfilter = REALLOC(excfilter, strlen(args)+1)) == NULL) {
+						OUT_OF_MEMORY /* LCOV_EXCL_LINE */
+					}
+					strcpy(excfilter, args);
+					filteropt = 1;
+				} else {
+					printf("Cannot filter and exclude simultaneously\n");
```
CurlyMoo: Can we make this a `logprinf` LOG_ERR.

receive.c
```diff
@@ -392,31 +468,60 @@ int main(int argc, char **argv) {
...
-					protocol = pnode->listener;
-					if(protocol_device_exists(protocol, filters[j]) == 0 && match == 0) {
-						match = 1;
-						break;
+		if(incfilter != NULL) {
+			struct protocol_t *protocol = NULL;
+			nrincfilter = explode(incfilter, ",", &incfilters);
+			unsigned int match = 0, j = 0;
+
+			protocol_init();
```
CurlyMoo: Why here?

receive.c
pilino1234: We can, will push new commit soon

receive.c
pilino1234: True, I'll move this and `struct protocol_t *protocol = NULL;` and `unsigned int match = 0, j = 0;` to just after `if(filteropt == 1) {`

pilino1234:
Updated, new code has a lot less duplication

CurlyMoo:
Can you make filtertype a integer. Then you give the states a label through a `#define`.

pilino1234:
Ah, that makes everything simpler. Will do

receive.c
```diff
@@ -109,10 +112,37 @@ static void on_read(int fd, char *buf, ssize_t len, char **buf1, ssize_t *len1)
 				for(i=0;i<n;i++) {
 					struct JsonNode *json = json_decode(array[i]);
 					if(json != NULL) {
-						char *out = json_stringify(json, "\t");
-						printf("%s\n", out);
+						if(filtertype != 0) {
+							int filtered = 0, j = 0;
+							char *protocol = NULL;
+							if(filtertype == EXCLUDE) {
+								filtered = 0;
```
CurlyMoo: Whats the point of doing this here and also on line 127?

receive.c
```diff
@@ -121,10 +151,37 @@ static void on_read(int fd, char *buf, ssize_t len, char **buf1, ssize_t *len1)
 			} else {
 				struct JsonNode *json = json_decode(*buf1);
 				if(json != NULL) {
-					char *out = json_stringify(json, "\t");
-					printf("%s\n", out);
+					if(filtertype != 0) {
```
CurlyMoo: What about turning this stuff in a seperate function? That would deduplicate the whole even more.

receive.c
pilino1234: Because the condition for the include filtering doesn't work otherwise - it still prints everything. It would be clearer if it was 

```
if(filtertype == INCLUDE) {
        filtered = 1;
}
```

receive.c
pilino1234: Yeah, probably a good idea

receive.c
```diff
@@ -99,6 +102,40 @@ static void alloc_cb(uv_handle_t *handle, size_t len, uv_buf_t *buf) {
 	memset(buf->base, 0, len);
 }
 
+static void printfilter(JsonNode *json) {
```
CurlyMoo: I think `output` would be better. The filtering is optional.

receive.c
```diff
@@ -391,13 +436,14 @@ int main(int argc, char **argv) {
 		}
 	}
 
-	if(filteropt == 1) {
+	if(filtertype != 0) {
```
CurlyMoo: `> 0` is even more accurate.

pilino1234:
Updated

pilino1234:
squashed

receive.c
```diff
@@ -99,6 +102,40 @@ static void alloc_cb(uv_handle_t *handle, size_t len, uv_buf_t *buf) {
 	memset(buf->base, 0, len);
 }
 
+static void output(JsonNode *json) {
```
CurlyMoo: Can you make this `struct JsonNode *json`

receive.c
```diff
@@ -352,7 +384,8 @@ int main(int argc, char **argv) {
 				printf("\t -P --port=xxxx\t\t\tconnect to server port\n");
 				printf("\t -I --instance=name\t\tconnect to pilight instance\n");
 				printf("\t -s --stats\t\t\tshow CPU and RAM statistics\n");
-				printf("\t -F --filter=protocol\t\tfilter out protocol(s)\n");
+				printf("\t -F --filter=protocol\t\tfilter for protocol(s)\n");
+				printf("\t -X --exclude=protocol\t\texclude protocol(s) from output\n");
```
CurlyMoo: The description is not quite uniform. The `-X` says  `... from output`, while the `-F` does not.

receive.c
```diff
@@ -382,7 +415,19 @@ int main(int argc, char **argv) {
 					OUT_OF_MEMORY /*LCOV_EXCL_LINE*/
 				}
 				strcpy(filter, args);
-				filteropt = 1;
+				filtertype = INCLUDE;
+			break;
+			case 'X':
+				if(filtertype == 0) {
+					if((filter = REALLOC(filter, strlen(args)+1)) == NULL) {
```
CurlyMoo: A `MALLOC` is more self-documenting, because there is not much to *re*alloc.

receive.c
```diff
@@ -382,7 +415,19 @@ int main(int argc, char **argv) {
...
+						OUT_OF_MEMORY /* LCOV_EXCL_LINE */
+					}
+					strcpy(filter, args);
+					filtertype = EXCLUDE;
+				} else {
+					logprintf(LOG_ERR, "Cannot filter and exclude simultaneously\n");
```
CurlyMoo: `logprintf` always add a newline so this can be removed here.

receive.c
```diff
@@ -468,6 +514,11 @@ int main(int argc, char **argv) {
 		options = NULL;
 	}
 
+	if(filter != NULL) {
+		FREE(filter);
+		filter = NULL;
```
CurlyMoo: `FREE` already sets the variable to `NULL` so this can be removed.

pilino1234:
Updated, seems like the `proc` unittest fails occasionally, although unrelated to the diff.

receive.c
```diff
@@ -391,13 +436,14 @@ int main(int argc, char **argv) {
...
 
-	if(filteropt == 1) {
+	if(filtertype > 0) {
 		struct protocol_t *protocol = NULL;
-		nrfilter = explode(filter, ",", &filters);
 		unsigned int match = 0, j = 0;
 
 		protocol_init();
 
+		nrfilter = explode(filter, ",", &filters);
```
CurlyMoo: I miss the freeing uos the filters array? Have you checked with valgrind?

pilino1234: I did in fact check with valgrind and it showed up fine 😕… Thought I had this free'd, apparently I don't, but it still works. Wtf? 😕
Will fix though.

pilino1234: Found it, it uses the same free as the previous implementation: https://github.com/pilight/pilight/pull/367/files#diff-eb620f32a6e8e7ceb3963dd45fb70665R59

Somehow this was already present in the rewrite code, so it's not shown in the diff.

CurlyMoo: You're right :smile;

CurlyMoo:
Thanks!

PR #373, opened 2017-09-02
Fix typo in liblua library detection

pilino1234:


pilino1234:
Should future PRs of this calibre be directed at development or rewrite (or both)?

CurlyMoo:
Both

CurlyMoo:
So what's your oppinion on the lua stuff ❓

pilino1234:
Still figuring out how it all works, but I think it will make it a lot easier for people to code their own operators. I assume it will be expanded to include functions and actions (and possibly also protocols?) in the future :)

CurlyMoo:
Yes, those will be next, but the operators where most easy to port.

pilino1234:
Nice. I'm currently trying to figure out what the almighty luastate `L` is, and what calls like [`lua_pcall(L, 2, 1, 0)`](https://github.com/pilight/pilight/commit/ed2e1e310b5ecf357d9cda6f5b2e55842f0618b5#diff-e37548eb19cddeb96af46f8545e4056cR132) do 😆

PR #380, opened 2017-09-13
Restructure travis to ease addition of more configurations

pilino1234:
This gives a cleaner `.travis.yml`, making it easier to add more configurations such as other compilers and linting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,42 +4,16 @@ sudo: true
 
 cache: ccache
 
-env:
-  matrix:
-    - CONFIG=DEFAULT
-#    - CONFIG=ALL_ON
-#    - CONFIG=ALL_OFF
+matrix:
+  fast_finish: true
+  include:
 
-addons:
-  apt:
-    packages:
-      - libpcap-dev
-      - libunwind8-dev
-      - lcov
-      - liblua5.2-0
-      - liblua5.2-dev
+    - env: PLATFORM=Unix COMPILER=gcc COVERAGE=1
+      compiler: gcc
+      os: linux
 
-install:
-  - curl https://apt.pilight.org/pilight.key | sudo apt-key add - && echo "deb http://apt.pilight.org/ stable main" | sudo tee -a /etc/apt/sources.list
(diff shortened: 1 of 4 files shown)


ci/travis-script.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e
+
+if [[ ${PLATFORM} == "Unix" ]]; then
+    mkdir -p build
+    cd build || exit 1
+
+    cmake -DPILIGHT_UNITTEST=1 \
+        -DCOVERALLS=ON \
+        -DCMAKE_BUILD_TYPE=Debug ..
+    make -j2
```
CurlyMoo: Can you remove this dual core compilation?

.travis.yml
```diff
@@ -4,42 +4,16 @@ sudo: true
...
-  apt:
-    packages:
-      - libpcap-dev
-      - libunwind8-dev
-      - lcov
-      - liblua5.2-0
-      - liblua5.2-dev
+    - env: PLATFORM=Unix COMPILER=gcc COVERAGE=1
+      compiler: gcc
+      os: linux
```
CurlyMoo: Are there other platforms supported?

ci/travis-script.sh
pilino1234: Will do

.travis.yml
pilino1234: The other supported option is osx, which probably isn't necessary at this stage.

I was thinking about doing windows compilations too (they would need to be cross-compiled with mingw). However, I have no idea how to set up cross-compiling toolchains, so I left this out for now

CurlyMoo: Windows has it's own `travis` but i need to fix windows native compilation first (which i started on rewrite).

pilino1234: > Windows has it's own travis

which CI does/will it use?

CurlyMoo: It's called AppVeyor https://www.appveyor.com/

CurlyMoo:
Can you check the coveralls error in the latest travis builds:
https://travis-ci.org/pilight/pilight/builds/311483459

pilino1234:
Should be fixed by #403

PR #382, opened 2017-09-13
Add more compilers to the Travis build matrix

pilino1234:
This ensures that pilight compiles on a wide variety of systems and
configurations.
This includes the default compiler for debian jessie, as well as some newer
versions of both gcc and clang.


pilino1234:
Seems the unittests are extra touchy today…

PR #381, opened 2017-09-13
add support for setting the resolution of ds18b20 sensors

KoffeinKaio:
Hey,

I'm slowly getting in the pilight code and thought of something easy to get started.

Since Linux Kernel version ~4.7, the w1-therm driver supports setting the resolution of the ds18b20 sensors by writing to the dev file.

This change adds a 'resolution' setting to the protocol which allows the configuration for 9/10/11/12 bit resolution

PR #383, opened 2017-09-14
add support for setting the resolution of ds18b20 sensors

KoffeinKaio:
rewrite for rewrite (heh.)
tested on dev and rewritten for the rewrite branch

original PR #381 desc from dev:

> 
> I'm slowly getting in the pilight code and thought of something easy to get started.
> 
> Since Linux Kernel version ~4.7, the w1-therm driver supports setting the resolution of the ds18b20 sensors by writing to the dev file.
> 
> This change adds a 'resolution' setting to the protocol which allows the configuration for 9/10/11/12 bit resolution

### DIFF
--- a/libs/pilight/protocols/GPIO/ds18b20.c
+++ b/libs/pilight/protocols/GPIO/ds18b20.c
@@ -17,6 +17,7 @@
 #include <sys/stat.h>
 #ifndef _WIN32
 	#include <unistd.h>
+	#include <sys/utsname.h>
 	#ifdef __mips__
 		#define __USE_UNIX98
 	#endif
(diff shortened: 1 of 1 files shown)


CurlyMoo:
Can you include your changes in the unit test?

KoffeinKaio:
Yeah, worked on that.

So far I#ve added the new setting to the device add string and skipped the file write if the test is running.
I would like to test the file write too, but that would require creating a device file instead of a simple "text" file I think, otherwise it would just truncate the file.


Any thoughts?

CurlyMoo:
A device file would be even better indeed!

KoffeinKaio:
To my researches a device file (pipe/fifo) would require something like a thread or an extra program to write to it while the test reads because it blocks.
So far I've found nothing that fits my needs, like some sort of buffered fifo.
Maybe I'll just resolve to write to the end of the text file, should not cause problems

CurlyMoo:
You can check the lirc unit test, which uses a named socket for this. That could work.

PR #385, opened 2017-09-22
New approach to pulse processing

CurlyMoo:
This PR investigates if we can remove all pilight assumptions about protocol processing and can create a new core that can handle every scenario there is. This should both allow old protocols to work and new protocols like S*mfy and Oregon to be implemented properly as well.


CurlyMoo:
@wo-rasp @pilino1234 @ma-ca, what do you guys think?

PR #391, opened 2017-10-13
Added manual page for secudo protocol

cxxcoder:
I added a manual page for the secudo protocol and created a pull request.
I created it by using another manual page as template and changing things to the best of my knowledge.
However, I'm not sure about the config stuff contained in the manual page.
The smoke sensors simply send repeatedly a signal if they detect smoke or the test button is pressed which will generate repeatedly corresponding messages in pilight's JSON output with the code changes of commit 91ffcb3.
I personally do not use pilight's gui and have not defined anything for the smoke sensors in the pilight configuration because I only use pilight's JSON output.
I don't know if there is any reasonable output in the gui if one defines such a smoke sensor in pilight's configuration.

### DIFF
--- a/docs/source/protocols/433.92/alarm/secudo.rst
+++ b/docs/source/protocols/433.92/alarm/secudo.rst
@@ -1,0 +1,115 @@
+.. |yes| image:: ../../../images/yes.png
+.. |no| image:: ../../../images/no.png
+
+.. role:: underline
+   :class: underline
+
+Secudo
+======
+
++------------------+-------------+
+| **Feature**      | **Support** |
++------------------+-------------+
+| Sending          | |no|        |
++------------------+-------------+
+| Receiving        | |yes|       |
++------------------+-------------+
+| Config           | |yes|       |
++------------------+-------------+
+
+.. rubric:: Supported Brands
+
++-----------------------+----------------+
+| **Brand**             | **Protocol**   |
++-----------------------+----------------+
+| Secudo                | secudo         |
++-----------------------+----------------+
+| FlammEx               | secudo         |
++-----------------------+----------------+
+
+.. rubric:: Sender Arguments
+
+*None*
+
+.. rubric:: Config
+
+.. code-block:: json
+   :linenos:
+
+   {
+     "devices": {
+       "alarm": {
+         "protocol": [ "secudo" ],
+         "id": [{
+           "id": 493
+         }],
+         "state": "alarm"
+       }
+     }
+   }
+
++------------------+-----------------+
+| **Option**       | **Value**       |
++------------------+-----------------+
+| id               | 0 - 1023        |
++------------------+-----------------+
+| state            | alarm           |
++------------------+-----------------+
+
+.. rubric:: Optional Settings
+
+*None*
+
+.. rubric:: Protocol
+
+This protocol sends 26 pulses like this:
+
+.. code-block:: console
+
+   314 314 628 628 314 314 628 314 628 628 314 314 628 314 628 314 628 314 628 628 314 628 314 314 628 10676
+
+The first pulse is the ``header`` and the last pulse is the ``footer``.
+These are meant to identify the pulses as genuine.
+We don't use them for further processing.
+The next step is to transform the remaining pulses into 12 groups of two pulses (and thereby dropping the ``header`` and ``footer`` pulses):
+
+.. code-block:: guess
+
+   314 628
+   628 314
+   314 628
+   314 628
+   628 314
+   314 628
+   314 628
+   314 628
+   314 628
+   628 314
+   628 314
+   314 628
+
+If we now look carefully at these groups you can distinguish two types of groups:
+
+#. 314 628
+#. 628 314
+
+So the first group is defined by a high 2nd and the second group by a high 1st pulse.
+So we take either of these two pulses to define a 0 or a 1.
+In this case we say a high 2nd pulse means a 1 and a low 2nd pulse means a 0.
+We then get the following output:
+
+.. code-block:: guess
+
+   101101111001
+
+The first 10 bits correspond to the setting of the ten dip switches inside the smoke sensor.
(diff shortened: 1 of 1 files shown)


docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,72 @@
+.. |yes| image:: ../../../images/yes.png
+.. |no| image:: ../../../images/no.png
+
+.. role:: underline
+   :class: underline
+
```
CurlyMoo: Can you add a header here:
```
Secudo
======
```

docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,72 @@
...
++------------------+-------------+
+| Config           | |yes|       |
++------------------+-------------+
+
+.. rubric:: Supported Brands
+
++---------------------------+------------------+
+| **Brand**                 | **Protocol**     |
++---------------------------+------------------+
+| secudo (formerly FlammEx) | secudo           |
```
CurlyMoo: Can you do this here:
```
+-----------------------+----------------+
| **Brand**             | **Protocol**   |
+-----------------------+----------------+
| Secudo                | secudo         |
+-----------------------+----------------+
| FlammEx               | secudo         |
+-----------------------+----------------+
```

docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,72 @@
...
++---------------------------+------------------+
+
+.. rubric:: Config
+
+.. code-block:: json
+   :linenos:
+
+   {
```
CurlyMoo: This should be:
```
   {
     "devices": {
       "alarm": {
         "protocol": [ "secudo" ],
         "id": [{
           "id": 493
         }],
         "state": "alarm"
       }
     }
   }
```
Does it reset itself to a non-alarm state?

docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. rubric:: Protocol
+
+This protocol sends 26 pulses like this
+
+.. code-block:: console
+
+   248 341 558 651 279 310 620 310 620 651 279 310 620 310 620 310 589 341 589 651 279 651 279 310 620 10633
+
+The pairs of the 2nd and 3rd, the 4th and 5th etc. pulses are either long-short or short-long.
```
CurlyMoo: Can you describe the protocol more like this:
https://github.com/pilight/pilight/blob/development/docs/source/protocols/433.92/contact/kaku.rst

cxxcoder:
I've updated the pull request by a force push to keep history clean.
New commit is 9890f4.
There is no reset to a non-alarm state because the smoke sensors only send one type of signal in the alarm case but no special signal if the alarm stops.

docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,122 @@
...
+     "devices": {
+       "alarm": {
+         "protocol": [ "secudo" ],
+         "id": [{
+           "id": 493
+         }],
+         "state": "alarm"
+       }
+     },
+     "gui": {
```
CurlyMoo: The idea was also to remove the gui part.

docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. rubric:: Protocol
+
+This protocol sends 26 pulses like this:
+
+.. code-block:: console
+
+   314 314 628 628 314 314 628 314 628 628 314 314 628 314 628 314 628 314 628 628 314 628 314 314 628 10676
+
+The first pulse is the ``header`` and the last pulse is the ``footer``.
```
CurlyMoo: Is this true? I don't see a header?

cxxcoder:
I'm not an RF expert and do not have much knowledge about the internals of
an RF telegram.
I wrote this because the 2nd and 3rd pulse, the 4th and 5th pulse and so on
encode the bits and thus the first pulse does not seem to carry any
information and I wanted to describe it similar as the manual page for kaku
contact as suggested by you.
If you know it better, you are welcome to adapt it. It is OK for me if the
commit for the secudo manual page does not have my username.


2017-10-13 23:39 GMT+02:00 CurlyMoo <notifications@github.com>:

docs/source/protocols/433.92/alarm/secudo.rst
```diff
@@ -0,0 +1,122 @@
...
+   101101111001
+
+The first 10 bits correspond to the setting of the ten dip switches inside the smoke sensor.
+If a dip switch is ON, this corresponds to 1 and 0 otherwise.
+The meaning of the two last bits is unknown.
+When reversing the first 10 bits because the first bits are least signifant we get the smoke sensors ID as binary number
+
+.. code-block:: guess
+
+   0111101101
```
CurlyMoo: Shouldn't the zero here correspond to the dip switches? When i look at the zero's here i only see the 1st, 6th and 8th bit differently, not the 2nd, 5th and 10th?

cxxcoder:
You have to count from right to left because when writing numbers the least
significant places are on the right.
But if you think that this is misleading, you can of course delete the last
lines.

2017-10-14 11:37 GMT+02:00 CurlyMoo <notifications@github.com>:

docs/source/protocols/433.92/alarm/secudo.rst
CurlyMoo: Can you still remove the GUI part.

CurlyMoo:
> You have to count from right to left because when writing numbers the least significant places are on the right.

I misread that part in the doc. Only the GUI part left.

cxxcoder:
I removed the GUI part. New commit is a064ffd9.

PR #393, opened 2017-10-21
Add a wrapping div to weather values

sreinelt:
Adding a wrapping div allows the values to stay together in one row in case of a line break on small screens (i.e. smartphones).

Before:
grafik

After:
grafik


CurlyMoo:
Can you do this with the other devices types as well (although the issue may not easily occur there)?

sreinelt:
Except for the DateTime element, I don't think that a wrapper div makes much sense, because there is nothing that can get torn apart by a line break.
I've only switches and weather in my setup. I'll add some generic devices to see how the other controls behave on my phone. But don't expect any progress on this in the next 3 or 4 weeks as my coming weekends are completely booked.

CurlyMoo:
> Except for the DateTime element, I don't think that a wrapper div makes much sense, because there is nothing that can get torn apart by a line break.

They don't do now, but no-one knows what happens in the future, and if all elements are consistent, we can make sure the future us aren't going to make the same mistakes.

CurlyMoo:
Does this fix the issue of @NiekD?

NiekD:
@sreinelt 
Thanks, this makes viewing on my phone a lot nicer!
I wonder if you could do this also for label values. I have some  that are (much) too long to be displayed on my phone screen.

NiekD:
@sreinelt, I am sorry, my request is in the wrong place. The label values now do not wrap at all, so this not relevant here.

CurlyMoo:
Can you resolve the issue?

sreinelt:
There is no issue, but I'm not allowed to remove the conflict marker.

CurlyMoo:
There is a conflict with `libs/webgui/pilight.css` that only you can solve.

PR #388, opened 2017-10-02
Break long switch caption in multiple lines

sreinelt:
This is a small css modification to allow long captions of switches to break into multiple lines.

snap_2017 10 02 10 43 51_001

PR #394, opened 2017-11-12
Make webgui take 'stats-enable' setting into account

pilino1234:
Previously, if stats were disabled, the webgui would still show the proc section (`CPU …%, RAM …%), with the placeholder always being empty. Fixed in this PR by taking the stats-enable setting into account when creating the proc section in the header of the webgui.


libs/webgui/pilight.js
```diff
@@ -12,6 +12,7 @@ var aMarqueueInterval = new Array();
 var aReadOnly = new Array();
 var aStates = new Array();
 var bShowTabs = true;
+var bStatsEnabled;
```
CurlyMoo: Can you initialize it

libs/webgui/pilight.js
```diff
@@ -1321,6 +1322,11 @@ $(document).ready(function() {
 			} else {
 				startWebsockets();
 			}
+			if('settings' in data && 'stats-enable' in data['settings']) {
+				console.log("setting foung")
```
CurlyMoo: debug typo's

libs/webgui/pilight.js
```diff
@@ -1321,6 +1322,11 @@ $(document).ready(function() {
...
+				bStatsEnabled = data['settings']['stats-enable'] == 1
+				console.log(bStatsEnabled)
```
CurlyMoo: likewise

libs/webgui/pilight.js
pilino1234: oops

pilino1234:
Fixed.
I based this on the current development branch, but that isn't based on the current master :/

CurlyMoo:
Nope. Staging is the new development.

pilino1234:
Changing target branch to `staging` was a bad idea 😂
Should I rebase this onto `staging` locally?

CurlyMoo:
I introduced a staging branch and made development a `feature` branch.

pilino1234:
Alright, that sounds good! Updated this to work with the correct branches now 😄

libs/webgui/pilight.js
```diff
@@ -1297,6 +1298,9 @@ $(document).ready(function() {
 			} else {
 				startWebsockets();
 			}
+			if('settings' in data && 'stats-enable' in data['settings']) {
+				bStatsEnabled = data['settings']['stats-enable'] == 1
```
CurlyMoo: I don't think the `== 1` is necessary.

pilino1234: Hmm, it seems it's unnecessary. Just wanted to be sure about truthy and false-y values in js first

pilino1234:
Tested and works, should be ready for merge :wink:

libs/webgui/pilight.js
```diff
@@ -1297,6 +1298,9 @@ $(document).ready(function() {
 			} else {
 				startWebsockets();
 			}
+			if('settings' in data && 'stats-enable' in data['settings']) {
+				bStatsEnabled = data['settings']['stats-enable']
```
CurlyMoo: Semicolon

pilino1234:
Fixed

CurlyMoo:
What did you change?

pilino1234:
Added the requested semicolon

CurlyMoo:
How did that fix the layout issue 😲

pilino1234:
That's a different discussion? 😛

CurlyMoo:
Conclusion?

pilino1234:
Dunno know much about the issue caused by #388. It seems very unrelated to this PR though, so I don't think I'll fix it here.

CurlyMoo:
You posted a screenshot earlier with an issue right?

pilino1234:
Nope, that was @NiekD here: https://github.com/pilight/pilight/commit/892289638dfa4edb6934e4c05a6bcb3ca16ad3ec#commitcomment-25607780 😛

CurlyMoo:
Aha.

pilino1234:
This should be good for merge though 😜

CurlyMoo:
And does it support both Ajax and websockets?

pilino1234:
Yes, ajax webgui doesn't actually contain the proc info in the first place, so no hiding to be done here

CurlyMoo:
Can you fix this for `development`, i will port to staging and master.

pilino1234:
Fixed, although it's a bit confusing that `development` isn't based on `master` currently...

CurlyMoo:
That's the idea of a `feature` branch. As soon as all code in development has been ported to master i will remove development and switch to feature branches.

PR #397, opened 2017-11-23
Break up long label text into multiple lines

NiekD:
Currently a label text, longer than 50% of the available width, is only partially visible.

With this commit a label text is allowed to use 95% of the available width and wrap around if it is even longer. 
This ensures that label text is always completely  visible, also on narrow screens.


CurlyMoo:
Can you post a screenshot of the previous and current situation?

NiekD:
Before:
image

After:
image

PR #402, opened 2017-12-06
Fix segfault in isrunning() on encountering long process name

pilino1234:
Running the slack desktop app while running the unittests revealed that the
process name was too long for the function to store in a char array of size
255. Fixed by increasing the size of the array to 1023, which seems to work
so far.

### DIFF
--- a/libs/pilight/core/common.c
+++ b/libs/pilight/core/common.c
@@ -234,11 +234,11 @@ int isrunning(const char *program) {
 		return -1;
 	}
 	int i = 0;
-	char name[255], *p = name;
-	memset(&name, '\0', 255);
+	char name[1024], *p = name;
+	memset(&name, '\0', sizeof(name));
 
 	for(i=0;i<psutil_max_pid();i++) {
-		if(psutil_proc_name(i, &p) == 0) {
+		if(psutil_proc_name(i, &p, sizeof(name)) == 0) {
 			if(strcmp(name, program) == 0) {
 				return i;
 			}
--- a/libs/pilight/psutil/linux.c
+++ b/libs/pilight/psutil/linux.c
@@ -43,9 +43,9 @@ int psutil_proc_exe(const pid_t pid, char **name) {
 	return -1;
 }
 
-int psutil_proc_name(const pid_t pid, char **name) {
+int psutil_proc_name(const pid_t pid, char **name, size_t bufsize) {
 	int ptr = 0, fd = 0;
-	char path[512], *p = path, cmdline[1024];
+	char path[512], *p = path, cmdline[bufsize];
 
 	snprintf(p, 511, "/proc/%d/cmdline", pid);
 	fd = open(path, O_RDONLY, 0);
@@ -53,7 +53,7 @@ int psutil_proc_name(const pid_t pid, char **name) {
 		return -1;
 	}
 
-	if((ptr = (int)read(fd, cmdline, 1023)) > 0) {
+	if((ptr = (int)read(fd, cmdline, bufsize-1)) > 0) {
 		strcpy(*name, basename(cmdline));
 		close(fd);
 		return 0;
--- a/libs/pilight/psutil/freebsd.c
+++ b/libs/pilight/psutil/freebsd.c
@@ -35,7 +35,7 @@ int psutil_proc_exe(const pid_t pid, char **name) {
 	return 0;
 }
 
-int psutil_proc_name(const pid_t pid, char **name) {
+int psutil_proc_name(const pid_t pid, char **name, size_t bufsize) {
 	// Fills a kinfo_proc struct based on process pid.
 	struct kinfo_proc proc;
 	int mib[4];
--- a/libs/pilight/psutil/psutil.h
+++ b/libs/pilight/psutil/psutil.h
@@ -4,6 +4,6 @@
 
 unsigned long psutil_max_pid(void);
 int psutil_proc_exe(const pid_t pid, char **name);
-int psutil_proc_name(const pid_t pid, char **name);
+int psutil_proc_name(const pid_t pid, char **name, size_t bufsize);
 int psutil_cpu_count_phys(void);
 size_t psutil_get_cmd_args(pid_t pid, char **args, size_t size);
--- a/libs/pilight/psutil/windows.c
+++ b/libs/pilight/psutil/windows.c
@@ -95,7 +95,7 @@ int psutil_proc_exe(const pid_t pid, char **name) {
 	return 0;
 }
 
-int psutil_proc_name(const pid_t pid, char **name) {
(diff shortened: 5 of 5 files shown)


pilino1234:
Backtrace when debugging revealed this:

```
(gdb) bt full
#0  0x00007ffff73f190c in isrunning (program=0x45ebad "pilight-unittest")
    at /home/martin/git/pilight/rewrite/libs/pilight/core/common.c:248
        i = 892678957
        name = "index.js --hidden-page --num-raster-threads=4 --enable-main-frame-before-activation --content-image-texture-target=0,0,3553;0,1,3553;0,2,3553;0,3,3553;0,4,3553;0,5,3553;0,6,3553;0,7,3553;0,8,3553;0,9,"...
        p = 0x7fffffffe7e0 "index.js --hidden-page --num-raster-threads=4 --enable-main-frame-before-activation --content-image-texture-target=0,0,3553;0,1,3553;0,2,3553;0,3,3553;0,4,3553;0,5,3553;0,6,3553;0,7,3553;0,8,3553;0,9,"...
```

Entering some of these command line options returned one slack helper process all the time. Terminating the Slack client stopped the segfault from happening. The full name of the process is this:

    martin   13571  0.2  1.5 1174036 123136 ?      Sl   11:20   0:01 /usr/lib/slack/slack --type=renderer --disable-pinch --no-sandbox --primordial-pipe-token=4AD2AD13EDBFBF49FCC41D140BEEBFF9 --lang=en-GB --standard-schemes=slack-resources,slack-webapp-dev --app-path=/usr/lib/slack/resources/app.asar --node-integration=true --preload=/usr/lib/slack/resources/app.asar/src/static/index.js --hidden-page --num-raster-threads=4 --enable-main-frame-before-activation --content-image-texture-target=0,0,3553;0,1,3553;0,2,3553;0,3,3553;0,4,3553;0,5,3553;0,6,3553;0,7,3553;0,8,3553;0,9,3553;0,10,3553;0,11,3553;0,12,3553;0,13,3553;0,14,3553;0,15,3553;1,0,3553;1,1,3553;1,2,3553;1,3,3553;1,4,3553;1,5,3553;1,6,3553;1,7,3553;1,8,3553;1,9,3553;1,10,3553;1,11,3553;1,12,3553;1,13,3553;1,14,3553;1,15,3553;2,0,3553;2,1,3553;2,2,3553;2,3,3553;2,4,3553;2,5,3553;2,6,3553;2,7,3553;2,8,3553;2,9,3553;2,10,3553;2,11,3553;2,12,3553;2,13,3553;2,14,3553;2,15,3553;3,0,3553;3,1,3553;3,2,3553;3,3,3553;3,4,3553;3,5,3553;3,6,3553;3,7,3553;3,8,3553;3,9,3553;3,10,3553;3,11,3553;3,12,3553;3,13,3553;3,14,3553;3,15,3553 --disable-accelerated-video-decode --disable-webrtc-hw-encoding --disable-gpu-compositing --service-request-channel-token=4AD2AD13EDBFBF49FCC41D140BEEBFF9 --renderer-client-id=3 --v8-natives-passed-by-fd --v8-snapshot-passed-by-fd

libs/pilight/core/common.c
```diff
@@ -234,8 +234,8 @@ int isrunning(const char *program) {
 		return -1;
 	}
 	int i = 0;
-	char name[255], *p = name;
-	memset(&name, '\0', 255);
+	char name[1023], *p = name;
+	memset(&name, '\0', 1023);
 
 	for(i=0;i<psutil_max_pid();i++) {
 		if(psutil_proc_name(i, &p) == 0) {
```
CurlyMoo: I want to suggest to make the buffer length a parameter of the `psutil_proc_name` function as well, because inside that function `read` is called with a buffer of 1024 as well. So, better make the whole stack dynamic:
https://github.com/pilino1234/pilight/blob/330addd253fcb4018f91f1e8278ff0f9af7b0933/libs/pilight/psutil/linux.c#L56

pilino1234:
I gave it a shot 😄

libs/pilight/core/common.c
```diff
@@ -234,11 +234,11 @@ int isrunning(const char *program) {
...
 	}
 	int i = 0;
-	char name[255], *p = name;
-	memset(&name, '\0', 255);
+	char name[1023], *p = name;
+	memset(&name, '\0', 1023);
 
 	for(i=0;i<psutil_max_pid();i++) {
-		if(psutil_proc_name(i, &p) == 0) {
+		if(psutil_proc_name(i, &p, 1024) == 0) {
```
CurlyMoo: I would say `sizeof(name)`. And there are also two other platforms in psutil.

CurlyMoo:
And did it work?

libs/pilight/core/common.c
pilino1234: Can't spot where to actually use the new parameter in the windows and freebsd code tho

pilino1234:
Compiles and the isrunning test passes for me

libs/pilight/core/common.c
CurlyMoo: Then the functions should still match.

pilino1234:
Squashed

pilino1234:
Thoughts on this PR? Would be good to fix, because it's quite annoying :/

CurlyMoo:
Christmas is the busiest time of the year 🎄

libs/pilight/core/common.c
```diff
@@ -234,11 +234,11 @@ int isrunning(const char *program) {
 		return -1;
 	}
 	int i = 0;
-	char name[255], *p = name;
-	memset(&name, '\0', 255);
+	char name[1023], *p = name;
```
CurlyMoo: Why not 1024?

libs/pilight/core/common.c
```diff
@@ -234,11 +234,11 @@ int isrunning(const char *program) {
...
+	memset(&name, '\0', 1023);
```
CurlyMoo: I would do `sizeof(name)` here as well.

libs/pilight/psutil/linux.c
```diff
@@ -43,17 +43,17 @@ int psutil_proc_exe(const pid_t pid, char **name) {
...
+	char path[512], *p = path, cmdline[bufsize];
 
 	snprintf(p, 511, "/proc/%d/cmdline", pid);
 	fd = open(path, O_RDONLY, 0);
 	if(fd == -1) {
 		return -1;
 	}
 
-	if((ptr = (int)read(fd, cmdline, 1023)) > 0) {
+	if((ptr = (int)read(fd, cmdline, bufsize-1)) > 0) {
```
CurlyMoo: Before of the 1023 before, this makes 1022.

pilino1234:
Better?

CurlyMoo:
Yes

PR #404, opened 2017-12-07
fixed systemd service

300481:
changed keyword "After" to "Requires" network-online.target
with only "After" this service will not start correctly after reboot

### DIFF
--- a/res/init/pilight.systemd
+++ b/res/init/pilight.systemd
@@ -1,6 +1,6 @@
 [Unit]
 Description=pilight
-After=network-online.target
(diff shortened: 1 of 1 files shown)


CurlyMoo:
H

300481:
okay, sorry. saw too late that i have to create a pull request against dev branch instead of master branch

CurlyMoo:
Yes, but you don't have to create a new one, but you can just change branches and fix the current one.

300481:
Next time! Thanks for fixing it ;-)

PR #406, opened 2017-12-16
logging: Always pad timestamp to same length (rewrite)

pilino1234:
This fixes inconsistent looking timestamps in the logging output by always
padding the microsecond component to 6 digits


pilino1234:
Note that this isn't as useful here, due to pilight logging the file and line number that the log messages originate at. How about only logging the file for warning and error messages?

CurlyMoo:
> How about only logging the file for warning and error messages?

I will think about it. However, that still has a challenge for the unit tests 👍

pilino1234:
So, which loglevels should log the file and line number? And (how) should this be tested?

CurlyMoo:
I will think about it again 🙄

CurlyMoo:
Any ideas?

PR #377, opened 2017-09-08
Astyle linting

pilino1234:
This allows checking the codestyle for stylistic errors and leads to a generally
nicer codebase.

The linting rules in util/astyle.conf are just preliminary, please give feedback
and requests on these 😄

PR #407, opened 2017-12-18
Add git pre-commit hook script

pilino1234:
The commit d5a5454 will disappear from this once #377 is merged


CurlyMoo:
This you fix this commit so it's actually a split.

pilino1234:
What's the new Codacy PR review thing that showed up? It looks quite nice IMO, very useful features!

CurlyMoo:
> What's the new Codacy PR review thing that showed up? It looks quite nice IMO, very useful features!

I was wondering what it was, but wasn't convinced enough.

pilino1234:
The static analysis features seem pretty nice, but that's about it.
I thought one could modify the Code Style settings, but apparently not. The rest of the metrics are fairly useless. Code Style, Error Prone, and Security seem to all point to exactly the same information. The interface is fancy, but after a closer look I'm not convinced by the features.

pilino1234:
Taking c71c4ca out of this will cause a merge conflict, because the modification to CONTRIBUTING.md in 0094072 (which this PR is about) is based on the changes in c71c4ca and the `astyle-linting` branch.
Rebasing this onto `rewrite` only to rebase it again after #377 is merged seems a bit pointless :/

CurlyMoo:
The PR's should be like this
1. Contribution and pre-commit hook
2. Astyle linting in travis.

The first is about setting the rules, second about checking if the rules were followered.

pilino1234:
Done!

ci/pre-commit
```diff
@@ -0,0 +1,23 @@
...
+if ! [ -x "$(command -v astyle)" ]; then
+    echo 'Error: astyle is not installed.' >&2
+    exit 1
+fi
+
+# Get path to the root of the repository
+GIT_DIR=`git rev-parse --show-toplevel`
+
+# Run the bundled astyle script
+$GIT_DIR/ci/lint.sh
```
pilino1234: Although this won't work currently, need to add ci/lint.sh to this PR too

CurlyMoo:
`$GIT_DIR/ci/lint.sh` shouldn't that go into the other PR?

pilino1234:
No, because then this PR wouldn't actually work, because there is not linter script to run from the hook.

CurlyMoo:
But our current CI is only about validating, not about changing code. So, should this file go into the CI folder in the first place?

pilino1234:
Hmm, after #377 is merged it should 😛
Currently, the pre-commit hook is in there too. Would this be better:
```
pilight-source-dir/
├── clients/
├── inc/
├── …
└── util/
    ├── ci/
    │   └── travis-*
    ├── pre-commit
    └── lint.sh
```


That's why it was easier to have both parts (adding linter script+git hook, and adding this to travis) in the same PR to begin with, because they are quite closely related.

CurlyMoo:
What about my very first suggestion long ago. To put it all in the `res` folder?

CurlyMoo:
Any ideas?

PR #408, opened 2018-01-02
binary.c: fix handling of s=0 in BITS_MSB_FIRST_TO_VALUE()

puuu:
This fix a bug introduced with 5eb10328acfd740570d00825c33f656b9591cc86.

`s` can be 0  as by the coment of `binToDecRev()`: [`0<=s<=e`](https://github.com/pilight/pilight/blob/development/libs/pilight/core/binary.c#L54)

This affect the quigg_gt9000 protocol, as it calls [`binToDecRev()` with `s=0`](https://github.com/pilight/pilight/blob/development/libs/pilight/protocols/433.92/quigg_gt9000.c#L161)

Results of `binToDecRev(binToDecRev((int[]){1,1,1,1, 1,1,1,1}, 0, 7))`:
before 5eb10328acfd740570d00825c33f656b9591cc86: 255
after 5eb10328acfd740570d00825c33f656b9591cc86: 127
this patch: 255

PR #409, opened 2018-01-08
fix handling of s=0 in BITS_MSB_FIRST_TO_VALUE() for rewrite

puuu:
Alternative pr for #408 against rewrite branch and with unit tests.

The patch differs from #408 to make `BITS_MSB_FIRST_TO_VALUE()` compatible with unsigned `e`, as in `binToDecRevUl()`.

If some or all commits should be merge, please ask for it.


puuu:
Squashed into one single commit.

PR #412, opened 2018-02-21
Make pilight.service require and wait for network-online.target

kylex2:
Tested on Raspbian lite stretch.

### DIFF
--- a/gen_package.sh
+++ b/gen_package.sh
@@ -27,6 +27,7 @@ mkdir -p source/etc/systemd/system
 mkdir -p source/etc/init
 cp ../res/init/pilight.sysv source/etc/init.d/pilight
 cp ../res/init/pilight.systemd source/etc/systemd/system/pilight.service
+cp ../res/init/wait-for-route.systemd source/etc/systemd/system/wait-for-route.service
 cp ../res/init/pilight.upstart source/etc/init/pilight.conf
 cp ../res/pilight.pem source/etc/pilight/pilight.pem
 
--- a/res/deb/pilight/postinst
+++ b/res/deb/pilight/postinst
@@ -1,4 +1,5 @@
 #!/bin/bash
+test -f /etc/os-release && . /etc/os-release
 
 if [[ $1 == "configure" ]]; then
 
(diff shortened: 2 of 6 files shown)


CurlyMoo:
Can you explain in your own words what this fixes? Also changed the title to something more meaningfull.

kylex2:
This is intended to fix the issue that pilight starts before it can bind to the multicast network.
`pilight-daemon: ERROR: cannot bind to the ssdp multicast network`

wait-for-route.service:
This is to make the network-online target in systemd work properly on Raspbian. If your system uses a network manager such as NetworkManager or similar it usually provides a service for a properly delayed network-online.target. Raspbian as default doesn't and that's what wait-for-route.service is trying to solve. It will delay the network-online target until a default route has been set or timeout after 2 minutes.

pilight.service:
The added Wants=network-online.target tells systemd that the service actually needs network-online.target to be brought up. Without that and no other service requiring that target it won't even be brought up. After= and Before= seems to more be used of ordering startup of services than what the service actually needs. 

Here are some quotes from the systemd manual that can be of interest for this one:
"If a unit foo.service contains a setting Before=bar.service and both units are being started, bar.service's start-up is delayed until foo.service has finished starting up. Note that this setting is independent of and orthogonal to the requirement dependencies as configured by Requires=, Wants= or BindsTo=."

Wants="A weaker version of Requires=. Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole. This is the recommended way to hook start-up of one unit to the start-up of another unit."

"Often, it is a better choice to use Wants= instead of Requires= in order to achieve a system that is more robust when dealing with failing services."

CurlyMoo:
What happens if the `res/init/wait-for-route.systemd` already exists?

kylex2:
Do you mean what if anyone created a service with the same name as me and then tries to install pilight?

CurlyMoo:
Yes

CurlyMoo:
I would still love to merge this PR, but i'm stilling waiting for that final answers.

CurlyMoo:
I renamed the whole thing by adding a prefix `pilight-` in 257318fd

PR #400, opened 2017-11-25
Added (optional) blink feature to generic_label

NiekD:
Generic_label devices with the BLINK option configured, can be set
to blinking/not blinking  with the label action.
This commit includes:
- generic_label protocol  with BLINK option + documentation
- label action with BLINK option + documentation
- pilight.css with additions defining blink properties
- pilight.js with additions to reflect blinking/not blinking state in the webgui

PR #413, opened 2018-02-26
fixed segfault occuring with long device strings

NiekD:
In devices.c a fixed size buffer was being used to hold device strings. If a device string was too long to fit in the buffer a segfault could occur.

The buffer for device strings has now been changed from fixed- to dynamic size.
The buffer is initialized with 255 bytes like before, but more space is now allocated in case this is not sufficient.

### DIFF
--- a/libs/pilight/config/devices.c
+++ b/libs/pilight/config/devices.c
@@ -81,7 +81,8 @@ int devices_update(char *protoname, JsonNode *json, enum origin_t origin, JsonNo
 	/* Do we need to update the devices file */
 	unsigned short update = 0;
 	/* The new state value */
-	char vstring_[255];
+	int vsize = 255;
+	char *vstring_ = MALLOC(vsize);
 	double vnumber_ = -1;
 	int vdecimals_ = 0;
 	char sstring_[255];
@@ -92,9 +93,13 @@ int devices_update(char *protoname, JsonNode *json, enum origin_t origin, JsonNo
 	/* The UUID of this device */
 	char *uuid = NULL;
 
+	if(vstring_ == NULL) {
+		fprintf(stderr, "out of memory\n");
+		exit(EXIT_FAILURE);
+	}
+
 	/* Make sure the character pointers are empty */
 	memset(sstring_, '\0', sizeof(sstring_));
-	memset(vstring_, '\0', sizeof(vstring_));
 
 	/* Check if the found settings matches the send code */
 	unsigned int match = 0, match1 = 0, match2 = 0;
@@ -255,9 +260,15 @@ int devices_update(char *protoname, JsonNode *json, enum origin_t origin, JsonNo
 									if(strcmp(sptr->name, opt->name) == 0
 									   && (opt->conftype == DEVICES_VALUE || opt->conftype == DEVICES_OPTIONAL)
 									   && opt->argtype == OPTION_HAS_VALUE) {
-										memset(vstring_, '\0', sizeof(vstring_));
 										vnumber_ = -1;
 										if(json_find_string(message, opt->name, &stmp) == 0) {
+											if((strlen(stmp) + 1) > vsize) {
+												vsize = strlen(stmp) + 1;
+												if((vstring_ = REALLOC(vstring_, vsize)) == NULL) {
+													fprintf(stderr, "out of memory\n");
+													exit(EXIT_FAILURE);
+												}
+											}
 											strcpy(vstring_, stmp);
 											valueType = JSON_STRING;
 											is_valid = 1;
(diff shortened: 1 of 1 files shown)


libs/pilight/config/devices.c
```diff
@@ -81,7 +81,8 @@ int devices_update(char *protoname, JsonNode *json, enum origin_t origin, JsonNo
 	/* Do we need to update the devices file */
 	unsigned short update = 0;
 	/* The new state value */
-	char vstring_[255];
+	int vsize = 255;
+	char *vstring_ = MALLOC(vsize);
```
CurlyMoo: Please check if the MALLOC failed.

libs/pilight/config/devices.c
```diff
@@ -255,9 +255,15 @@ int devices_update(char *protoname, JsonNode *json, enum origin_t origin, JsonNo
 									if(strcmp(sptr->name, opt->name) == 0
 									   && (opt->conftype == DEVICES_VALUE || opt->conftype == DEVICES_OPTIONAL)
 									   && opt->argtype == OPTION_HAS_VALUE) {
-										memset(vstring_, '\0', sizeof(vstring_));
 										vnumber_ = -1;
 										if(json_find_string(message, opt->name, &stmp) == 0) {
+ 											if(strlen(stmp) + 1 > vsize) {
```
CurlyMoo: Can you increase the size in larger steps to prevent constant increasing of the size.

libs/pilight/config/devices.c
NiekD: OK, I will

libs/pilight/config/devices.c
```diff
@@ -308,11 +314,17 @@ int devices_update(char *protoname, JsonNode *json, enum origin_t origin, JsonNo
 								   && (opt->conftype == DEVICES_VALUE || opt->conftype == DEVICES_OPTIONAL)
 								   && opt->argtype == OPTION_HAS_VALUE) {
 									int upd_value = 1;
-									memset(vstring_, '\0', sizeof(vstring_));
 									vnumber_ = -1;
 									vdecimals_ = 0;
 									struct JsonNode *jtmp = NULL;
 									if(json_find_string(message, opt->name, &stmp) == 0) {
+										if((strlen(stmp) + 1) > vsize) {
```
CurlyMoo: You didn't enclose this in a hook above.

libs/pilight/config/devices.c
NiekD: I doubt if that is really necessary, because the initial size is 255. Only in rare cases strings will be longer than 254 and the chance of having more than one of such strings in one device is very small.

libs/pilight/config/devices.c
NiekD: Yes indeed, that's not consistent.

libs/pilight/config/devices.c
CurlyMoo: Increasing the size by just 1 is almost always a bad idea.

NiekD: But then I'm afraid that you didn't really see what I did. 

I am not incrementing by 1. The new allocated size is always the length of the string that doesn't fit  (plus 1 for the terminator ofc).  The +1 you are pointing at is in the if statement, not in the realloc.

CurlyMoo: You are right. Sorry. But, then increase by a fixed number above the length of the max string. E.g. steps of 255.

NiekD: Ofc I can do that, but imho then we are needlesly waisting memory, because the chance that another string of the same device is longer than the one that makes it necessary to enlarge the buffer, is almost zero.

For instance a label text can be very long, but the other strings (color and in my case also blink and bgcolor) are never that long and will always fit. And the same goes for the http protocol: the data returned can be very long, but the other strings (mimetype and state) never are. And I can't think of any feasable protocol that would have multiple strings longer than 254 chars.

NiekD:
Yes, it is. Without the fix my pilight always crashes when the length of a label text exceeds the buffer size. 

Although the chance that this will happen in practice is rather small, this applies to any device string getting longer than the buffer size.

CurlyMoo:
Do you know with what commit?

NiekD:
As far as I know this has been an issue from the beginning, but was not discovered before, because long label text was useless before the introduction of wrapping long label text.

CurlyMoo:
Sorry, the "With what commit" was meant for the `str_replace` PR. 

What do we need to do to get this merged?

NiekD:
I guess bringing my fork up to date with staging to resolve the conflicts should be enough to make this PR ready to be merged.

NiekD:
I have updated my fork, but realized that my PRs are all in separate branches taken from a version of the development branch that was current at the time I did the PRs, but is now much outdated. 
Because of the many changes since then (a.o. introduction of lua for actions and functions and all related library- and directory changes), this causes many conflicts that all need to be manually resolved for each PR separately before I can even get to updating/migrating the modules involved with the PR's.

I tried to do it, but without success.  Git even seems to get into problems and comes up with errors not referring to any file in the repo. After many hours of trying, I am fed up with it now and want to spend my time on more productive things.
 
Therefor I would suggest to delete my existing fork, create a fresh new one and make new clean PRs for the current staging branch, with a comment referring to the corresponding old PRs.
The old PRs can then be closed with a final comment referring to the new ones.

That way we are not losing the history of discussions and comments and can still make a fresh start.

CurlyMoo:
Maybe you can give me access to your forks and let me fix things. I believe i'm a little more handy with git :smile:

NiekD:
That would be great!

My repo's are public. The branches for the PRs:

[#400: https://github.com/NiekD/pilight/tree/NiekD-label-blink](https://github.com/NiekD/pilight/tree/NiekD-label-blink)

[#413: https://github.com/NiekD/pilight/tree/devices-fix-overflow](https://github.com/NiekD/pilight/tree/devices-fix-overflow)

[#414: https://github.com/NiekD/pilight/tree/additional-unittest-for-str_replace ](https://github.com/NiekD/pilight/tree/additional-unittest-for-str_replace)

CurlyMoo:
Updated

CurlyMoo:
All done

NiekD:
Imho all comments have been handled before. 
So, now that all conflicts have been solved too, I think this PR can be merged.

CurlyMoo:
This one fails:
https://travis-ci.org/pilight/pilight/builds/474365020?utm_source=github_status&utm_medium=notification

PR #414, opened 2018-03-11
Additional unittest for str_replace

NiekD:
Added two tests demonstrating a bug in str_replace, causing a wrong new length.
This occurs if:

- The replacement string is longer than the search string and
- The offset of the search string found in the target string is lower than the length of the replacement string minus the length of the search string.

### DIFF
--- a/tests/common.c
+++ b/tests/common.c
@@ -515,6 +515,16 @@ static void test_str_replace(CuTest *tc) {
 	CuAssertIntEquals(tc, 19, n);
 	CuAssertStrEquals(tc, "<body text='black'>", str);
 
+	strcpy(str, "Apple");
+	n = str_replace("Apple", "Pineapple", &p);
+	CuAssertIntEquals(tc, 9, n);
+	CuAssertStrEquals(tc, "Pineapple", str);
+
+	strcpy(str, "My fruit");
+	n = str_replace("fruit", "raspberry", &p);
+	CuAssertIntEquals(tc, 12, n);
+	CuAssertStrEquals(tc, "My raspberry", str);
+
 	CuAssertIntEquals(tc, 0, xfree());
 }
 
--- a/libs/pilight/core/common.c
+++ b/libs/pilight/core/common.c
@@ -795,11 +795,10 @@ int str_replace(char *search, char *replace, char **str) {
 			match = 1;
 			int rpos = (x + (slen - rlen));
 			if(rpos < 0) {
-				slen -= rpos;
 				rpos = 0;
 			}
 			nlen = len - (slen - rlen);
-			if(len < nlen) {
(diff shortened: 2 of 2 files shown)


CurlyMoo:
Can you check (and add to your PR the code of) this patch:
```
diff --git a/libs/pilight/core/common.c b/libs/pilight/core/common.c
index 83e63cb8..58678564 100644
--- a/libs/pilight/core/common.c
+++ b/libs/pilight/core/common.c
@@ -791,11 +791,10 @@ int str_replace(char *search, char *replace, char **str) {
                        match = 1;
                        int rpos = (x + (slen - rlen));
                        if(rpos < 0) {
-                               slen -= rpos;
                                rpos = 0;
                        }
                        nlen = len - (slen - rlen);
-                       if(len < nlen) {
+                       if(slen < rlen) {
                                if(((*str) = REALLOC((*str), (size_t)nlen+1)) == NULL) { /*LCOV_EXCL_LINE*/
                                }
```

NiekD:
I applied the patch and tested it succesfully. 
I don't think however that after removing the "slen -= rpos" line, it is really necessary to change the "if" for the realloc, because "len < nlen" and "slen < rlen" now are in fact identical.

CurlyMoo:
> because "len < nlen" and "slen < rlen" now are in fact identical.

The second is more self-documenting.

CurlyMoo:
This should be discussed fully right?

NiekD:
The patch you proposed was fixing the issue and you said preferred the modified "if" condition for a good reason.
Is there more to this we should discuss?

CurlyMoo:
Yes, but my idea was never pushed. I though you were going to include it in your PR.

CurlyMoo:
Ok, so are you going to do it? That would prevent another PR not merged.

NiekD:
Ok. After bringing my fork up to date with staging I will add the fix to this PR, so you can merge it.

NiekD:
It fails on the str_replace, which is in a different PR (#414), and the reported issue is not related to PR #413

It ought to fail without the patch. I must investigate why the test still fails with the patched string_replace.

tests/common.c
```diff
@@ -515,6 +515,16 @@ static void test_str_replace(CuTest *tc) {
 	CuAssertIntEquals(tc, 19, n);
 	CuAssertStrEquals(tc, "<body text='black'>", str);
 
+	strcpy(str, "Apple");
+	n = str_replace("Apple", "Pineapple", &p);
```
CurlyMoo: Now i'm actually checking this test, i don't see how this should result into `Pineapple`?

CurlyMoo: My mistake. It should.

CurlyMoo:
I fixed it.

PR #427, opened 2019-01-05
fix build for arm

mwka:


CurlyMoo:
Fixing things does not work by just commenting out stuf :smile:

mwka:
That is true - should I copy the part from the staging branch which is using wiringXSupportedPlatforms()?

CurlyMoo:
If that works, yes.

PR #432, opened 2019-01-23
new 433 Mhz protocol: HIDEKI

f-n:
The supported weather station is called "Wetterstation Meteotime" and was provided by the German Aldi Supermarket. The external temperature/humidity sensor was analysed by the tool "rtl_433": The tool detected the protocol as "hideki". This implementation uses information from the rtl_433 implementation for this protocol: it also supports wind and rain sensors: Although the code exists in my changes, only the temperature/humidity sensor could be tested.


CurlyMoo:
Please read the contribution guide.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
+/*
+	Copyright (C) 2018
+
+	This file is part of pilight.
```
CurlyMoo: Please make this MPLv2

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+
+#define MIN_PULSE_LENGTH			220
+#define MAX_PULSE_LENGTH			280
+
+#define MIN_RAW_LENGTH			130
+#define MAX_RAW_LENGTH		180
+#define PULSE_HIDEKI_WEATHER_LOWER		100	// experimentally detected
+#define PULSE_HIDEKI_WEATHER_UPPER		700	// experimentally detected
+
+#define HIDEKI_MAX_BYTES_PER_ROW 14
```
CurlyMoo: Check indenting here.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+ *    SYNC+HEAD cha   RC   B LEN        Nr.?   RAIN_L    RAIN_H     0x66       XOR       CRC
+ *
+ */
+
+enum sensortypes { HIDEKI_UNKNOWN, HIDEKI_TEMP, HIDEKI_TS04, HIDEKI_WIND, HIDEKI_RAIN };
+
+typedef struct settings_t {
+	/* This driver implements different sensor types, that are dynamically detected in parseCode() function.
+	 * Unfortunately, the detected sensor type sometimes is wrong. Unfortunately, the expected sensor type
+	 * as defined in the user configuration ("protocol" name from config.jaon) is not accessible in function
```
CurlyMoo: Type: `jaon`

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+	 * checkValues(). Hence a heuristics tries to get the right sensor type and stores them in this settings
+	 * structure:
+	 * In the first received packets, the system stores the sensor type as soon as the same type was detected
+	 * in (CALIBRATION_COUNT+1) subsequent data packets. From this time on, all decoded packets with
+	 * wrong sensor type are ignored */
+#define CALIBRATION_COUNT  3
```
CurlyMoo: Please move this to the top.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+	double wind_dir;
+	double wind_factor;
+	double rain;
+	struct settings_t *next;
+} settings_t;
+
+static struct settings_t *settings = NULL;
+
+static uint8_t reverse8(uint8_t x)
+{
```
CurlyMoo: Consistent style please

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+
+static int byteParity(uint8_t inByte)
+{
+    inByte ^= inByte >> 4;
+    inByte &= 0xf;
+    return (0x6996 >> inByte) & 1;
+}
+
+static void set_bit(uint8_t *array, int position) {
+	uint8_t mask = 1 << (7 - (position % 8));
```
CurlyMoo: Indenting

CurlyMoo: Or check the other indenting if this one is correct.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+
+static int validate(void) {
+	int x = 0, i = 0;
+	uint8_t header = 0;
+
+	if ((hideki->rawlen < MIN_RAW_LENGTH) || (hideki->rawlen > MAX_RAW_LENGTH))
+	  return -1;
+
+	for(x=0; x<8; x++) {
+		if(hideki->raw[i] > PULSE_HIDEKI_WEATHER_LOWER && 	hideki->raw[i] < PULSE_HIDEKI_WEATHER_UPPER) {
```
CurlyMoo: Bogus spaces

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+				set_bit(&header, x);
+				i++;
+			} else {
```
CurlyMoo: Just remove the else.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+				 // leave value unchanged to 0
+			}
+			i++;
+		}
+#ifdef HIDEKI_DEBUG
+	logprintf(LOG_DEBUG, "HIDEKI validate(): rawlen=%d, header byte: %02X", hideki->rawlen, header);
+#endif
+
+	if (header != 0x06) // tested with TS04 only
```
CurlyMoo: Style guide

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+static void parseCode(void) {
+	uint8_t binary[MAX_RAW_LENGTH/2];
+#ifdef HIDEKI_DEBUG
+	int id = 0;
+#endif
+	int i = 0, x;
+	double humi_offset = 0.0, temp_offset = 0.0, wind_dir_offset = 0.0, wind_factor = 1.0, rain_factor = 1.0;
+	uint8_t packet[HIDEKI_MAX_BYTES_PER_ROW];
+	int sensortype = HIDEKI_WIND; // default for 14 valid bytes
+    uint8_t channel, rc, battery_ok;
```
CurlyMoo: Indenting

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+    double temperature, humidity, wind_strength, wind_direction, rain;
+    uint8_t ignore_temperature = 0;
+
+	// Decode Biphase Mark Coded Differential Manchester (BMCDM) pulse stream into binary
+	memset(&binary[0], 0, sizeof(binary));
+	for(x=0; x<=(MAX_RAW_LENGTH/2); x++) {
+		if(hideki->raw[i] > PULSE_HIDEKI_WEATHER_LOWER && hideki->raw[i] < PULSE_HIDEKI_WEATHER_UPPER) {
+			set_bit(binary, x);
+			i++;
+		} else {
```
CurlyMoo: Remove the else.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+	//  * invert all bits
+	//  * Remove (and check) parity bit
+	for (i = 0; i < HIDEKI_MAX_BYTES_PER_ROW; i++) {
+	    unsigned int offset = i/8;
+	    packet[i] = (binary[i+offset] << (i%8)) | (binary[i+offset+1] >> (8 - i%8)); // skip/remove parity bit
+	    packet[i] = reverse8(packet[i]); // reverse LSB first to LSB last
+	    packet[i] ^= 0xFF; // invert bits
+	    // check parity
+	    uint8_t parity = ((binary[i+offset+1] >> (7 - i%8)) ^ 0xFF) & 0x01;
+	    if (parity != byteParity(packet[i]))
```
CurlyMoo: Indenting and style.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+
+#ifdef HIDEKI_DEBUG
+     logprintf(LOG_DEBUG, "HIDEKI parseCode(): channel %02X, ID:  %02X, RC: %02X, temp: %02X, batt: %02X, sensortyp: %02x, hum: %02X",
+			channel, id, rc, temp, battery_ok, sensortype, ((packet[6] & 0xF0) >> 4) * 10 + (packet[6] & 0x0F));
+#endif
+
+     // read some settings
+ 	 struct settings_t *my_settings = get_settings((double)channel);
+ 	 if (my_settings) {
+ 			humi_offset = my_settings->humi;
```
CurlyMoo: Indenting

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+ 			temp_offset = my_settings->temp;
+ 			wind_dir_offset = my_settings->wind_dir;
+ 			wind_factor = my_settings->wind_factor;
+ 			rain_factor = my_settings->rain;
+ 			if ((battery_ok == 0) && (my_settings->ignore_temp))
```
CurlyMoo: Style

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+ 				ignore_temperature = 1; // workaround for case that battery=0: temperature value might be invalid in that case
+
+ 			if  (my_settings->calibration > 0) {
```
CurlyMoo: Style

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+ 							sensortype, my_settings->sensortype, sensortype);
+ 				return;	// ignore the data
+ 			}
+  	 } else {
+			/* no valid settings/configuration found -> the received data do not belong to a configured device from config.json
+			 * -> usually we could return here, but in that case the tool "pilight_receive" might be unable to receive other data
+			 * than locally configured... */
+  	 }
+
+     if (sensortype == HIDEKI_TS04) {
```
CurlyMoo: Indenting

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+    	  temperature = (double)temp/10 + temp_offset;
+          humidity = ((packet[6] & 0xF0) >> 4) * 10 + (packet[6] & 0x0F) + humi_offset;
+
+          hideki->message = json_mkobject();
+          if (ignore_temperature == 0)
```
CurlyMoo: Style

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+         	  json_append_member(hideki->message, "temperature", json_mknumber(temperature, 1));
+          json_append_member(hideki->message, "humidity", json_mknumber(humidity, 0));
+
+          json_append_member(hideki->message, "battery", json_mknumber(battery_ok, 0));
+          json_append_member(hideki->message, "channel", json_mknumber(channel, 0));
+          json_append_member(hideki->message, "rc", json_mknumber(rc, 0));
+
+      } else  if (sensortype == HIDEKI_WIND) {
```
CurlyMoo: Bogus space

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+	options_add(&hideki->options, "c", "channel", OPTION_HAS_VALUE, DEVICES_ID, JSON_NUMBER, NULL, "[0-9]");
+	// device settings:
+	options_add(&hideki->options, "0", "temperature-offset", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)0, "[0-9]");
+	options_add(&hideki->options, "0", "humidity-offset", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)0, "[0-9]");
+	options_add(&hideki->options, "0", "wind-direction-offset", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)0, "[0-9]");
+	options_add(&hideki->options, "0", "wind-factor", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)1, "^[0-9]{1,5}$");
+	options_add(&hideki->options, "0", "rain-factor", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)1, "^[0-9]{1,5}$");
+	/* tests have shown that the transmitted temperature value might be damages in case if battery=0.
+	 * As workaround, the temperature will not be output in case of battery=0: This behavior can be activated by this setting */
+	options_add(&hideki->options, "0", "ignore-temperature-on-battery-low", OPTION_NO_VALUE, DEVICES_SETTING, JSON_NUMBER, NULL, NULL);
```
CurlyMoo: Why not make this the default?

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,509 @@
...
+	// GUI  settings:
```
CurlyMoo: Bogus spaces

libs/pilight/protocols/433.92/hideki.h
```diff
@@ -0,0 +1,27 @@
+/*
+	Copyright (C) 2018
+
```
CurlyMoo: License to MPLv2

libs/pilight/protocols/433.92/hideki.c
CurlyMoo: New PR's and new code is slowly moving to MPLv2 because i believe GPLv3 is too restrictive. Also, i don't like how GPL forces me to make everything GPL.

libs/pilight/protocols/433.92/hideki.c
f-n: I thought about this. For me it's a kind of workaround and I don't know the reason of this behaviour. I could test the sensor only with one single device. Maybe it's specific for my device or the location of the sensor. As long as I have no indication that this is a general problem of this sensors I prefer to leave it optional and switched off per default.

CurlyMoo: Was this protocol already discussed on the forum. Maybe other users can tune in?

f-n: the only location in the forum I remember is this one:
https://forum.pilight.org/showthread.php?tid=3421

libs/pilight/protocols/433.92/hideki.c
f-n: Do you have a MPLv2 header with copyright notice which I can use as template?

libs/pilight/protocols/433.92/hideki.c
f-n: So, I adjusted the formatting on various locations. Do I need to repeat the pull request for these changes?

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+ *    00000000  11111111  22222222  33333333  44444444  55555555  66666666  77777777  88888888 99999999 AAAAAAAA BBBBBBBB CCCCCCCC DDDDDDDD
+ *    SYNC+HEAD cha   RC     LEN        Nr.?    1° .1°  VB   10°    1° .1°  VB   10°   1W .1W  .1G 10W   10G 1G    w°  AA    XOR      CRC
+ *
+ * Rain:
+ *    00000000  11111111  22222222  33333333  44444444  55555555  66666666  77777777  88888888
+ *    SYNC+HEAD cha   RC   B LEN        Nr.?   RAIN_L    RAIN_H     0x66       XOR       CRC
+ *
+ */
+
+enum sensortypes { HIDEKI_UNKNOWN, HIDEKI_TEMP, HIDEKI_TS04, HIDEKI_WIND, HIDEKI_RAIN };
```
CurlyMoo: Can you make this static?

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+static void set_bit(uint8_t *array, int position) {
+	uint8_t mask = 1 << (7 - (position % 8));
+	array[position / 8] |= mask;
+}
+
+static struct settings_t *get_settings(double channel) {
+	struct settings_t *tmp = settings;
+
+	while(tmp) {
+		if( /*  fabs(tmp->rc - rc) < EPSILON) &&  */ fabs(tmp->channel - channel) < EPSILON)  {
```
CurlyMoo: Can you add a comment here about why you disabled this?

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+	}
+	return NULL;
+}
+
+static int validate(void) {
+	int x = 0, i = 0;
+	uint8_t header = 0;
+
+	if((hideki->rawlen < MIN_RAW_LENGTH) || (hideki->rawlen > MAX_RAW_LENGTH))
+	  return -1;
```
CurlyMoo: Indentation. And please encapsulate these in curly brackets as well.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+
+	for(x=0; x<8; x++) {
+		if(hideki->raw[i] > PULSE_HIDEKI_WEATHER_LOWER && hideki->raw[i] < PULSE_HIDEKI_WEATHER_UPPER) {
+				set_bit(&header, x);
+				i++;
+			}
```
CurlyMoo: Indentation

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+	//  * Remove (and check) parity bit
+	for(i = 0; i < HIDEKI_MAX_BYTES_PER_ROW; i++) {
+	    unsigned int offset = i/8;
+	    packet[i] = (binary[i+offset] << (i%8)) | (binary[i+offset+1] >> (8 - i%8)); // skip/remove parity bit
+	    packet[i] = reverse8(packet[i]); // reverse LSB first to LSB last
+	    packet[i] ^= 0xFF; // invert bits
+	    // check parity
+	    uint8_t parity = ((binary[i+offset+1] >> (7 - i%8)) ^ 0xFF) & 0x01;
+	    if(parity != byteParity(packet[i]))
+	    {
```
CurlyMoo: Inconsistent style

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+	    	if(i == 10) {
+	    		sensortype = HIDEKI_TS04;
+	            break;
+	         }
```
CurlyMoo: Indentation

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+	    unsigned int offset = i/8;
+	    packet[i] = (binary[i+offset] << (i%8)) | (binary[i+offset+1] >> (8 - i%8)); // skip/remove parity bit
+	    packet[i] = reverse8(packet[i]); // reverse LSB first to LSB last
+	    packet[i] ^= 0xFF; // invert bits
+	    // check parity
+	    uint8_t parity = ((binary[i+offset+1] >> (7 - i%8)) ^ 0xFF) & 0x01;
+	    if(parity != byteParity(packet[i]))
+	    {
+	    	if(i == 10) {
+	    		sensortype = HIDEKI_TS04;
```
CurlyMoo: Indentation

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+	// read some settings
+	struct settings_t *my_settings = get_settings((double)channel);
+	if(my_settings) {
+		humi_offset = my_settings->humi;
+ 		temp_offset = my_settings->temp;
+ 		wind_dir_offset = my_settings->wind_dir;
+ 		wind_factor = my_settings->wind_factor;
+ 		rain_factor = my_settings->rain;
+ 		
+		if((battery_ok == 0) && (my_settings->ignore_temp))
```
CurlyMoo: Here as well

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+		 * -> usually we could return here, but in that case the tool "pilight_receive" might be unable to receive other data
+		 * than locally configured... */
+  	}
+	
+	if(sensortype == HIDEKI_TS04) {
+		temperature = (double)temp/10 + temp_offset;
+		humidity = ((packet[6] & 0xF0) >> 4) * 10 + (packet[6] & 0x0F) + humi_offset;
+
+		hideki->message = json_mkobject();
+		if(ignore_temperature == 0)
```
CurlyMoo: And here

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+		const uint8_t wd[] = { 0, 15, 13, 14, 9, 10, 12, 11, 1, 2, 4, 3, 8, 7, 5, 6 };
+		temperature = (double)temp/10 + temp_offset;
+		wind_strength = (double)(packet[9] & 0x0F) * 100 + ((packet[8] & 0xF0) >> 4) * 10 + (packet[8] & 0x0F); // unit still unclear
+		wind_strength *= wind_factor; // user defined adjustment
+		// wind strength in km/h = wind_strength*0.160934
+		wind_direction = (double)(wd[((packet[11] & 0xF0) >> 4)] * 225);
+		wind_direction /= 10;
+		wind_direction += wind_dir_offset; // user defined adjustment
+#if 0
+		if(wind_direction > 360.0)
```
CurlyMoo: And here

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,492 @@
...
+			snode->wind_factor = 1.0;
+			snode->rain = 1.0;
+			snode->ignore_temp = 0;
+
+			json_find_number(jvalues, "temperature-offset", &snode->temp);
+			json_find_number(jvalues, "humidity-offset", &snode->humi);
+			json_find_number(jvalues, "wind-direction-offset", &snode->wind_dir);
+			json_find_number(jvalues, "wind-factor", &snode->wind_factor);
+			json_find_number(jvalues, "rain-factor", &snode->rain);
+			if(json_find_number(jvalues, "ignore-temperature-on-battery-low", &tmp) == 0)
```
CurlyMoo: And here

libs/pilight/protocols/433.92/hideki.c
f-n: why that? 
I just tries it out and the compiler responded with:
/usr/src/pilight/libs/pilight/protocols/433.92/hideki.c:66:82: warning: useless storage class specifier in empty declaration
 static enum sensortypes { HIDEKI_UNKNOWN, HIDEKI_TEMP, HIDEKI_TS04, HIDEKI_WIND, HIDEKI_RAIN };

CurlyMoo: Oeps, enums are always static. My bad.

libs/pilight/protocols/433.92/hideki.c
f-n: i will remove it: it was just for the case that the channel alone is not sufficient to identify the current device.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,497 @@
...
+static void set_bit(uint8_t *array, int position) {
+	uint8_t mask = 1 << (7 - (position % 8));
+	array[position / 8] |= mask;
+}
+
+static struct settings_t *get_settings(double channel) {
+	struct settings_t *tmp = settings;
+
+	while(tmp) {
+		if(fabs(tmp->channel - channel) < EPSILON)  {
```
CurlyMoo: Isn't it easier to cast channel to an int below so we can just compare the numbers directly without having to rely on the `fabs` and `EPSILON`?

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,497 @@
...
+			i++;
+		}
+		i++;
+	}
+#ifdef HIDEKI_DEBUG
+	logprintf(LOG_DEBUG, "HIDEKI validate(): rawlen=%d, header byte: %02X", hideki->rawlen, header);
+#endif
+
+	if(header != 0x06) {	// tested with TS04 only
+        	return -1;
```
CurlyMoo: Indenting

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,497 @@
...
+
+	return 0;
+}
+
+static void parseCode(void) {
+	uint8_t binary[MAX_RAW_LENGTH/2];
+#ifdef HIDEKI_DEBUG
+	int id = 0;
+#endif
+	int i = 0, x;
```
CurlyMoo: Can you tidy this up a little, and initialize all variables:
```
	double temperature = 0.0, humidity = 0.0, wind_strength = 0.0, wind_direction = 0.0, rain = 0.0;
	double humi_offset = 0.0, temp_offset = 0.0, wind_dir_offset = 0.0, wind_factor = 1.0, rain_factor = 1.0;
	int i = 0, x = 0;
	int temp = 0 rain_units = 0;
	int sensortype = HIDEKI_WIND; // default for 14 valid bytes
	uint8_t packet[HIDEKI_MAX_BYTES_PER_ROW];
	uint8_t channel = 0, rc = 0, battery_ok = 0;
	uint8_t ignore_temperature = 0;
	uint8_t binary[MAX_RAW_LENGTH/2];
```

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,497 @@
...
+			/* calibration is finished and detected sensor type does not match */
+ 			logprintf(LOG_NOTICE, "HIDEKI parseCode(): wrong sensor type found: %d (expected: %d) -> ignoring data",
+				  sensortype, my_settings->sensortype, sensortype);
+			return;	// ignore the data
+ 		}
+	} else {
+		/* no valid settings/configuration found -> the received data do not belong to a configured device from config.json
+		 * -> usually we could return here, but in that case the tool "pilight_receive" might be unable to receive other data
+		 * than locally configured... */
+  	}
```
CurlyMoo: Indenting

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,497 @@
...
+		hideki->message = json_mkobject();
+		if(ignore_temperature == 0) {
+			json_append_member(hideki->message, "temperature", json_mknumber(temperature, 1));
+		}
+		json_append_member(hideki->message, "humidity", json_mknumber(humidity, 0));
+
+		json_append_member(hideki->message, "battery", json_mknumber(battery_ok, 0));
+		json_append_member(hideki->message, "channel", json_mknumber(channel, 0));
+		json_append_member(hideki->message, "rc", json_mknumber(rc, 0));
+      } else if(sensortype == HIDEKI_WIND) {
```
CurlyMoo: You're pretty inconsistent in using spaces and tabs. That's why the indenting is wrong here using a tabular view.

f-n: I don't use an editor which shows the differences between spaces and tabs, so you are right. For me personally, this is not a problem.

CurlyMoo: I use notepad++ with a tabs = 2 spaces. And indeed it anoys me :smile:.

CurlyMoo: The other `sensortype` conditions also contain spaces.

libs/pilight/protocols/433.92/hideki.c
f-n: I agree with you. As I checked the corresponding code in other pilight protocol implementations I found that this is done in the same way. The I decided to leave it using double.

CurlyMoo: Maybe i was too tired back then. So let's not copy the same wierd implementations over and over.

f-n: OK, I will think about it.... but not today.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -380,7 +381,16 @@ static int checkValues(struct JsonNode *jvalues) {
 			jchild = jchild->next;
 		}
 
-		if(get_settings(channel) == NULL) {
+		if(channel < 0.0) {
```
CurlyMoo: If you configured the options right, these checks are redundant.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,512 @@
...
+	options_add(&hideki->options, "0", "wind-factor", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)1, "^[0-9]{1,5}$");
+	options_add(&hideki->options, "0", "rain-factor", OPTION_HAS_VALUE, DEVICES_SETTING, JSON_NUMBER, (void *)1, "^[0-9]{1,5}$");
+	/* tests have shown that the transmitted temperature value might be damages in case if battery=0.
+	 * As workaround, the temperature will not be output in case of battery=0: This behavior can be activated by this setting */
+	options_add(&hideki->options, "0", "ignore-temperature-on-battery-low", OPTION_NO_VALUE, DEVICES_SETTING, JSON_NUMBER, NULL, NULL);
+	// GUI  settings:
+	options_add(&hideki->options, "0", "temperature-decimals", OPTION_HAS_VALUE, GUI_SETTING, JSON_NUMBER, (void *)1, "[0-9]");
+	options_add(&hideki->options, "0", "humidity-decimals", OPTION_HAS_VALUE, GUI_SETTING, JSON_NUMBER, (void *)0, "[0-9]");
+	options_add(&hideki->options, "0", "wind-decimals", OPTION_HAS_VALUE, GUI_SETTING, JSON_NUMBER, (void *)1, "[0-9]");
+	options_add(&hideki->options, "0", "wind_direction-decimals", OPTION_HAS_VALUE, GUI_SETTING, JSON_NUMBER, (void *)0, "[0-9]");
```
CurlyMoo: Why does this setting have an underscore?

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,512 @@
...
+#if 0
+		if(wind_direction > 360.0) {
+			wind_direction -= 360.0;
+		}
+#endif
+		hideki->message = json_mkobject();
+		if(ignore_temperature == 0) {
+			json_append_member(hideki->message, "temperature", json_mknumber(temperature, 1));
+		}
+		json_append_member(hideki->message, "wind strength", json_mknumber(wind_strength, 1));
```
CurlyMoo: The wind options doesn't have spaces.

libs/pilight/protocols/433.92/hideki.c
```diff
@@ -0,0 +1,512 @@
...
+}
+
+#if !defined(MODULE) && !defined(_WIN32)
+__attribute__((weak))
+#endif
+void hidekiInit(void) {
+	protocol_register(&hideki);
+	protocol_set_id(hideki, "hideki");
+	protocol_device_add(hideki, "TS04", "HIDEKI TS04 sensor");
+	protocol_device_add(hideki, "Hideki-Wind", "HIDEKI Wind sensor");
```
CurlyMoo: Above you wrote that this is not something we can use inside a pilight protocol. Why still define it here?

libs/pilight/protocols/433.92/hideki.c
f-n: Why not? If you prefer a dash instead, I can/will replace it.

libs/pilight/protocols/433.92/hideki.c
f-n: Oh, I thought this argument is considered as human readable string. I can replace the space with dash, ok?

libs/pilight/protocols/433.92/hideki.c
f-n: if you refer to the code below the statement "if(parity != byteParity(packet[i]))" then I want to note that the sensortype variable is initialized with HIDEKI_WIND. This value stays valid in case that for all iterations of the for-loop the check "parity != byteParity(packet[i]))" is false.

CurlyMoo: But that piece of code isn't affected in any way by the `protocol_device_add` pieces. I would just remove those aliasses.

f-n: Ok, that was not so clear to me. In that case I'm wondering why other protocol implementations also use protocol_device_add() multiple times, e.g. tfa.c has 5 calls of protocol_device_add(). I think, multiple aliases can enable the user to setup the corresponding string into the config file to improve the readability of the config file. From my point of view it would be nice to get access to these settings from inside the checkValues() function. In that case pilight could know the sensortype that the user had configured.

f-n: I would like to know your opinion to the following issue: Since I did not identify a unique ID for the sensors, I basically use the channel number to distinguish the devices. My test devices can be configured only for 3 different channels. So, theoretically I can only have 3 sensors of that type. On the other hand, the sensores transmit a "rolling code", that could be used together with the channel number to distinguish the sensors. But there are disadvantages: The value of the rc changes on every battery change or reset of the sensor and the user has no influence to its value. After each battery change the rc canges to an unpredictable value and the user can only read the value by pilight-receive or something similar. So, a change of the config file would be necessary. What is your opinion: Does it make sense to include the rc value into the "id" section of the config file?

CurlyMoo: > protocol_device_add() multiple times

Because the same protocol can be implemented in different brands. In the future i want to remove those aliasses and just call them 'switch1', 'switch2', 'switch3' etc. And the same for the sensors, weather, dimmers etc.

> I think, multiple aliases can enable the user to setup the corresponding string into the config file to improve the readability of the config file

True

> In that case pilight could know the sensortype that the user had configured.

Why would that be usefull. Why not just have one configured device that stores all those values?

> Does it make sense to include the rc value into the "id" section of the config file?

Depends how much the batteries needs changed? If that's once every let's say 6 months, i would include it. If it's once each month, i would exclude it. One idea that comes up right now is the option of allowing a wild card. So that the device reacts to each channel or id.

f-n: > Why would that be usefull. Why not just have one configured device that stores all those values?

Because in my case the sensor type can be an additional criteria (id) to identify the device. So, devices of different types can share the same channel. 
> One idea that comes up right now is the option of allowing a wild card

This is a good idea. I will think about it: maybe the rolling code can be included into the id section and will only be used as criteria if the configured value is >=0

CurlyMoo: > Because in my case the sensor type can be an additional criteria (id) to identify the device. So, devices of different types can share the same channel.

Are those wind and rain meters just the same hardware device?

> This is a good idea.

At the moment however, this is still not possible.

f-n: I have no idea. I have never tested wind or rain meters.

CurlyMoo: Shall we just assume for now that those are the same devices and let the protocol communicate it that way?

f-n: Ok.

wo-rasp: > 
> 
> I would like to know your opinion to the following issue: Since I did not identify a unique ID for the sensors, I basically use the channel number to distinguish the devices. My test devices can be configured only for 3 different channels. So, theoretically I can only have 3 sensors of that type. On the other hand, the sensores transmit a "rolling code", that could be used together with the channel number to distinguish the sensors. But there are disadvantages: The value of the rc changes on every battery change or reset of the sensor and the user has no influence to its value. After each battery change the rc canges to an unpredictable value and the user can only read the value by pilight-receive or something similar. So, a change of the config file would be necessary. What is your opinion: Does it make sense to include the rc value into the "id" section of the config file?

PR #434, opened 2019-02-16
support for raspyrfm hardware with unittest

Phunkafizer:


tests/hardware_raspyrfm.c
```diff
@@ -0,0 +1,321 @@
...
+
+	uv_thread_join(&pth);
+
+	storage_gc();
+	plua_gc();
+	hardware_gc();
+	eventpool_gc();
+	wiringXGC();
+
+	CuAssertTrue(gtc, check == 1);
```
CurlyMoo: Can this become:
```
CuAssertIntEquals(gtc, check, 1);
```

tests/hardware_raspyrfm.c
```diff
@@ -0,0 +1,321 @@
...
+	json_delete(jsettings);
+}
+
+static void accept_gpio(uv_stream_t *req, int status) {
+	CuAssertTrue(gtc, status >= 0);
+
+	client_req = MALLOC(sizeof(uv_pipe_t));
+	CuAssertPtrNotNull(gtc, client_req);
+
+	int r = uv_pipe_init(uv_default_loop(), client_req, 0);
```
CurlyMoo: You aren't checking the `r` output.

tests/hardware_raspyrfm.c
```diff
@@ -0,0 +1,321 @@
...
+	eventpool_init(EVENTPOOL_THREADED);
+
+	if (start(__FUNCTION__, connect_send_cb) == -1)
+		return;
+}
+
+CuSuite *suite_hardware_raspyrfm(void) {
+	CuSuite *suite = CuSuiteNew();
+
+	SUITE_ADD_TEST(suite, test_hardware_raspyrfm_send);
```
CurlyMoo: Can you also add the tests that check the config parameters. As is check for valid and invalid parameters.

Phunkafizer:
While doing some tests concerning the topic https://forum.pilight.org/showthread.php?tid=3522 I got an issue; the hwtype was set too late so that pilight can handle multiple hardware. I forced push and will bring it also to the staging branch.

libs/pilight/hardware/raspyrfm.c
```diff
@@ -0,0 +1,441 @@
...
+#endif
+void raspyrfmInit(void) {
+	hardware_register(&raspyrfm);
+	hardware_set_id(raspyrfm, "raspyrfm");
+	
+	options_add(&raspyrfm->options, "f", "frequency", OPTION_HAS_VALUE, DEVICES_VALUE, JSON_NUMBER, NULL, "^[0-9.]+$");
+	options_add(&raspyrfm->options, "c", "spi-channel", OPTION_HAS_VALUE, DEVICES_VALUE, JSON_NUMBER, NULL, "^[01]$");
+
+	int frequency = 0;
+	if(config_setting_get_number("frequency", 0, &frequency) != 0) {
```
CurlyMoo: Don't forget to handle the error or actually enforce the options with the option mask.

CurlyMoo:
> I forced push and will bring it also to the staging branch.

What actually changed?

Phunkafizer:
> 
> 
> > I forced push and will bring it also to the staging branch.
> 
> What actually changed?

Also here, the setting of hwtype

libs/pilight/hardware/raspyrfm.c
```diff
@@ -0,0 +1,441 @@
...
+	}
+	
+	raspyrfm->minrawlen = 1000;
+	raspyrfm->maxrawlen = 0;
+	raspyrfm->mingaplen = 5100;
+	raspyrfm->maxgaplen = 10000;
+	
+	raspyrfm->comtype=COMOOK;
```
CurlyMoo: Also, check the style guide here

libs/pilight/hardware/raspyrfm.c
Phunkafizer: I missed some lines when pushing. Now frequency is checked.

CurlyMoo:
I'm currently porting the hardware module to lua, but i noticed that the unittest doesn't yet cover everything. 

What i countered till this far:
- It doesn't test if all necessary `rfmcfg` values are set to the correct values
- It doesn't test if the SPI device is opened with the correct speed.

CurlyMoo:
What i also don't get is how you set a rawlen of 32 pulses on line 202, while you expect a rawlen of 50 pulses on line 114.

Phunkafizer:
> 
> 
> What i also don't get is how you set a rawlen of 32 pulses on line 202, while you expect a rawlen of 50 pulses on line 114.

This is because the rfm69 module transforms pulselengths to a contineous bitstream. For every 31 µS a bit is generated. A value of 248 in txtestpulses generates 248/31 = 8 bits. The last value in txtestpulses (5952) results in 5952/31 = 192 bits to be generated, in this case represented by the last 24 bytes in txtestraw array.

Phunkafizer:
> 
> 
> I'm currently porting the hardware module to lua, but i noticed that the unittest doesn't yet cover everything.
> 
> What i countered till this far:
> 
>     * It doesn't test if all necessary `rfmcfg` values are set to the correct values
> 
>     * It doesn't test if the SPI device is opened with the correct speed.

How can I test for correct speed? I think if the SPI can not be opened with the requested speed it should return an error.

Which test is missing? The frequency is checked in setFrequency, if SPI channel is not provided, raspyrfmHwInit would return an error.

CurlyMoo:
> Which test is missing?

Here various parameters are set, but the test doesn't check if they are all set with the correct values:
https://github.com/pilight/pilight/pull/434/files#diff-37f44c5e034922929cf9d82ace01b336R323

> The frequency is checked in setFrequency

This one is set.

> How can I test for correct speed? I think if the SPI can not be opened with the requested speed it should return an error.

This is what i did in the new serial unittest:
```c
int tcgetattr(int fildes, struct termios *termios_p) {
	memset(termios_p, 0, sizeof(struct termios));
	return 0;
}

int tcsetattr(int fildes, int optional_actions, const struct termios *termios_p) {
	char buffer[1024];
	int l = 0;

	l += snprintf(buffer, 1024,
		"%4u",
		optional_actions);
	l += snprintf(&buffer[l], 1024-l,
		"%8u",
		termios_p->c_iflag);
	l += snprintf(&buffer[l], 1024-l,
		"%8u",
		termios_p->c_oflag);
	l += snprintf(&buffer[l], 1024-l,
		"%8u",
		termios_p->c_cflag);
	l += snprintf(&buffer[l], 1024-l,
		"%8u",
		termios_p->c_lflag);

	int i = 0;
	for(i=0;i<NCCS;i++) {
		l += snprintf(&buffer[l], 1024-l,
			"%4u",
			termios_p->c_cc[i]);
	}

	l += snprintf(&buffer[l], 1024-l,
		"%8u",
		termios_p->c_ispeed);
	l += snprintf(&buffer[l], 1024-l,
		"%8u",
		termios_p->c_ospeed);
	write(fildes, buffer, l);
	return 0;
}
```

And the other side:
```c
static void callback(uv_fs_t *req) {
	unsigned int len = req->result;

	uv_fs_req_cleanup(req);
	FREE(req);

	if(len > 0) {
		if(strcmp(rbuffer, "   2       0       0    6321       0   0   0   0   0   0 150   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0    4097    4097") == 0) {
			pthread_attr_t attrs;
			pthread_attr_init(&attrs);
			pthread_attr_setdetachstate(&attrs, PTHREAD_CREATE_DETACHED);

			pthread_create(&pthread, NULL, thread, &fd);
			return;
		}
	}

	if(run == 1) {
		uv_fs_t *read_req = NULL;
		if((read_req = MALLOC(sizeof(uv_fs_t))) == NULL) {
		}

		memset(rbuffer, '\0', BUFFER_SIZE);

		uv_buf_t buf = uv_buf_init(rbuffer, BUFFER_SIZE);
		uv_fs_read(uv_default_loop(), read_req, fd, &buf, 1, -1, callback);
	}
}
```

So, you could change the wiringXSPISetup to write the SPI parameters to the /dev/spidev0.0 file as just file content and then you check the parameters written to that file in a file read callback at the unittest side. That's at least what i did.

Phunkafizer:
> Here various parameters are set, but the test doesn't check if they are all set with the correct values:
> https://github.com/pilight/pilight/pull/434/files#diff-37f44c5e034922929cf9d82ace01b336R323
The rfmcfg are hardcoded values for the RFM69 IC. Is it really needed to check for it? When the SPI is working, which is checked implicitely with RX/TX tests, it can not fail.

CurlyMoo:
> The rfmcfg are hardcoded values for the RFM69 IC. Is it really needed to check for it? When the SPI is working, which is checked implicitely with RX/TX tests, it can not fail.

How can i be sure about that when porint the whole thing to lua and not having the device itself.

Phunkafizer:
> How can i be sure about that when porint the whole thing to lua and not having the device itself.

Good point. I added checking of rfmcfg.

CurlyMoo:
So what happens when i do the following:
```
for(i=0;i<27;i++) {
   writeReg(i+1, i+1);
}
```
instead of
```
for(i=0; i<sizeof(rfmcfg) / sizeof(rfmcfg[0]); i++)
    writeReg(rfmcfg[i].reg, rfmcfg[i].val);
```
Does the unittest work or fail and why?

Phunkafizer:
> Does the unittest work or fail and why?

It will fail. You are writing wrong values in wrong registers. In the unittest all register writes are catched, and if a register/value pair there matches with rfmcfg, I set a flag in rfmcfgcheck. When everything is flagged in the rfmcfgcheck, the test will pass.

CurlyMoo:
But where are those values defined at  the unit test side? I only see that empty struct.

Phunkafizer:
Those values are defined here:

https://github.com/pilight/pilight/blob/cbd9a35536a10eb705fbd9f1b9b1c028f0f22a01/libs/pilight/hardware/raspyrfm.c#L84

CurlyMoo:
That's what I said. At the hardware side not at the unit test side. 

Also, the new implementation will be in lua so the c defined struct of the module can't be used anymore .

Phunkafizer:
> 
> That's what I said. At the hardware side not at the unit test side. Also, the new implementation will be in lua so the c defined struct of the module can't be used anymore .

Ok, this means the rfmcfg needs to be defined redundant also in the unittest. Then of course if a value in rfmcfg has to be changed it will also have to be changed in the unittest.

CurlyMoo:
> Ok, this means the rfmcfg needs to be defined redundant also in the unittest.

Welcome to the world of unittesting 😉

CurlyMoo:
I've ported the whole module to lua and all unittest succeed. You can check the work done in the `raspyrfm` repository. If the module doesn't work in real life, then the unittests are incomplete. I would suggest making sure the critical parts currently missing are covered in the final unittest.

I also believe nobody has ever got a more thorough review than you got. I see that as a compliment :smile:

There are a few things i also want to see fixed before i finally merge:
- The `minrawlen` and `maxrawlen` are set specificly to the values of the test. But in the final pilight implementation those are not the values used. mingaplen: `5100`, maxgaplen: `99999`, minrawlen: `25`, maxrawlen: `512`.
- The same applies to the rx and tx tests. Those are not pulses i've ever seen in public. I would like to see actual pulses of a real protocol. You can use my 433nano and 433gpio as a reference. The pulses used there are those of a KlikAanKlikUit switch.
- The parameter tests are also quite incomplete. Like missing values, wrong SPI channels, too many option set also those not defined etc.
- In regard to the settings. Can we change the allowed frequency values to `433` and `868`. I would also like to see the `spi-channel` renamed to `channel`.

Lastly:
- Please make sure the style remains consistent to the pilight style.

Phunkafizer:
I'm back after a busy period.
I checked out the raspyrfm branch and was able to compile. Should the daemon run in this branch? Or is it only for the unittest? At the moment the daemon is looping @100% CPU in the loop in daemon.c line 1619

CurlyMoo:
pilight programs are not supported in the rewrite branch. That's why i removed them in the latest commits.

CurlyMoo:
Also, welcome back 👍

PR #438, opened 2019-05-12
Add unit tests for iwds07 protocol

1technophile:
Unit testing based on data coming from :
https://forum.pilight.org/showthread.php?tid=2952


CurlyMoo:
Ok, why was this closed?

PR #441, opened 2019-10-26
Start pilight without forking and restart on failure when using systemd

dstolpmann:
This pull request starts pilight without forking when running as a systemd services. This is the recommended starting method according to [1]. It also enables restarting on failure [2], which restarts the pilight daemon in case of an unexpected error and thus increases the reliability of the system.

[1] https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=
[2] https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

### DIFF
--- a/res/init/pilight.systemd
+++ b/res/init/pilight.systemd
@@ -3,7 +3,8 @@ Description=pilight
 After=network-online.target
 
 [Service]
-ExecStart=/usr/local/sbin/pilight-daemon -F
+ExecStart=/usr/local/sbin/pilight-daemon
+Type=forking
 Restart=on-failure
 
 [Install]
--- a/res/init/pilight.systemd
+++ b/res/init/pilight.systemd
@@ -3,8 +3,8 @@ Description=pilight
 After=network-online.target
 
 [Service]
-ExecStart=/usr/local/sbin/pilight-daemon
-Type=forking
+ExecStart=/usr/local/sbin/pilight-daemon -F
+Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target


CurlyMoo:
> This is the recommended starting method according to [1]

I'm not exactly sure what points you are referring to but what i was looking for was this (as described in the forking description):
```
The parent process is expected to exit when start-up is complete and all communication channels are set up.
```

Just as described in this stackoverflow answer:
```
If the prompt returns but the service keeps running in the background (i.e. the service daemonizes itself on its own), then Type = forking is the right choice.
```
https://superuser.com/questions/1274901/systemd-forking-vs-simple/1274913

There are several issues of the `simple` type:
```
its communication channels should be installed before the service is started up
```
This isn't the case in pilight

```
as the service manager will immediately proceed starting follow-up units, right after creating the main service process, and before executing the service's binary
```
So, the process created before executing pilight itself.

```
simple services will report success even if the service's binary cannot be invoked successfully
```
This is also not always the case in regard to pilight.

dstolpmann:
If the service daemonizes itself (as it is the case for pilight), "Type=forking" is indeed the right option. However, I added the -F option to the "ExecStart" command, which runs pilight in the foreground. This makes pilight behave the way it is expected for "Type=simple", which is the default when not specifying the type.

This is also recommended in the StackExchange answer that you cited:
> Note: many service applications designed to be portable to many Unix-style systems may behave as b) [Type=forking] by default, but can be made to work like a) [Type=simple] by adding an option (usually described as "don't fork", "keep running in foreground", "don't daemonize" or similar). In that case, if the option has no other side effects, then adding the option and using the a)-type behavior would be preferable for systemd.

However, you are right about the problem with the starting of the follow-up units. This would require to use "Type=notify" and call the systemd API from pilight once all communication channels are set up.

The main intention of this pull request was to add the "Restart=on-failure" functionality. I was not sure wether it will work properly with the forking type, thus I changed it to simple, which I know is working. However, I just tested the restarting functionality with the forking type and I can confirm that it works. Should I create another pull request with just adding the restarting functionality?

CurlyMoo:
> However, I added the -F option to the "ExecStart" command, which runs pilight in the foreground.

I saw that indeed.

> This makes pilight behave the way it is expected for "Type=simple", which is the default when not specifying the type.

No, not exactly. Because that's not the only assumption the simple type does, as you are implying in the rest of your answer.

> Should I create another pull request with just adding the restarting functionality?

No, change the current one.

dstolpmann:
OK, I changed the service type back to forking.

CurlyMoo:
How did you test the failure?

dstolpmann:
I sent a SIGSEGV to the pilight-daemon using:
`killall pilight-daemon --signal=SIGSEGV`

PR #451, opened 2020-03-24
Updated MIN_PULSE_LENGTH for KaKu APIR-2150 support

tvandergeer:
Now supports KaKu/CoCo APIR-2150 outdoor motion sensors

I've tested with 2 KlikAanKlikUit (KaKu) APIR-2150 outdoor motion sensors. They were not recognised by the current protocol. Debugging the RAW signal learned that a lower MIN_PULSE_LENGTH value would fix this.

In a batch of 14 samples the value for arctech_motion->raw[arctech_motion->rawlen-1] was between 5655 and 5662. Dividing with the PULSE_DIV of 34 a value of 165 would suffice. I chose 160 to be a bit on the safe side.

### DIFF
--- a/libs/pilight/protocols/433.92/arctech_motion.c
+++ b/libs/pilight/protocols/433.92/arctech_motion.c
@@ -30,7 +30,7 @@
 #include "arctech_motion.h"
 
 #define PULSE_MULTIPLIER	3
-#define MIN_PULSE_LENGTH	250
+#define MIN_PULSE_LENGTH	160
 #define MAX_PULSE_LENGTH	284
 #define AVG_PULSE_LENGTH	279
 #define RAW_LENGTH				132


CurlyMoo:
Was this related to the outstanding issue from the forum?

tvandergeer:
I was not aware of an outstanding topic on the forum...

Do you mean this: https://forum.pilight.org/showthread.php?tid=1160 ?
It could be a a solution for that. They didn't post RAWs so can not tell for sure.

tvandergeer:
Ah, there is a some raw data in this post: https://forum.pilight.org/showthread.php?tid=1160&pid=24994#pid24994

I can't make up which captures are from the motion sensor, but I see quite a lot of readings where the last value is in the same range as my readings: ~5660

CurlyMoo:
Do you have the possibility to make a rewrite PR as well so i can have some unittests for this device?

tvandergeer:
Can you show me an example of this? I'd be happy to help

tvandergeer:
Ah, "rewrite" is the name of a branch. I've found the protocol tests too. Coming up!

PR #455, opened 2020-04-05
Added unit tests for Smartwares switch

LeoCal:
Implemented and executed 7 new unit tests for smartwares switch.

```
~/pilight_rewrite/build# ./pilight-unittest test_protocols_433
[ test_protocols_433_alecto_ws1700                 ]
[ test_protocols_433_alecto_wx500                  ]
[ test_protocols_433_arctech_contact               ]
[ test_protocols_433_arctech_motion                ]
[ test_protocols_433_GS-iwds07                     ]
[ test_protocols_433_arctech_dimmer                ]
[ test_protocols_433_arctech_dusk                  ]
[ test_protocols_433_livolo_switch                 ]
[ test_protocols_433_smartwares_switch             ]
[ test_protocols_433_kerui_D026                    ]
[ test_protocols_433_tfa2017                       ]
.

OK (1 test)
```


CurlyMoo:
You only set the input codes, not the output once.

Input is used for devices that can only be recieved like contacts. This switch can be both sent and receive. Check arctech_dimmer for an example of both codes set.

LeoCal:
> You only set the input codes, not the output once.
> 
> Input is used for devices that can only be recieved like contacts. This switch can be both sent and receive. Check arctech_dimmer for an example of both codes set.

Oh ok, I see - my bad, I should have looked at raw_t struct notification to understand that better.
Unfortunately though, as said earlier, I do not own a receiver, so I went ahead and adjusted the test to have check output codes only and I can confirm I see all tests passing. I see the same approach has been taken for livolo_switch commit, so I hope that's fine with you.

CurlyMoo:
> Fixed output codes definition and removed input

Don't remove the input, but make sure both input and output are present.

LeoCal:
> > Fixed output codes definition and removed input
> 
> Don't remove the input, but make sure both input and output are present.

Done

CurlyMoo:
If i have to redevelop the protocol, ensuring these tests to succeed is enough ensurance?

LeoCal:
I would say level of protection from unit tests is more or less on par with the other protocols (https://github.com/pilight/pilight/tree/rewrite/tests/protocols) and surely better than protocols lacking tests (arctech_switch). All in all, I'd say it's a good start that anyone could then extend.

CurlyMoo:
> surely better than protocols lacking tests (arctech_switch)

Which i own myself, so i don't bother about that one :smile:

CurlyMoo:
Can you add two additional unittests:
- One with another unit code
- One that intentionally fails

LeoCal:
Done.

PR #457, opened 2020-08-15
add "battery low" state to kerui protocol

mwalliczek:
Add battery state to kerui protocol


CurlyMoo:
Can you elaborate these changes?

mwalliczek:
The change is to detect and report the state "low battery" that could be send by a kerui sensor device when the battery is low.

CurlyMoo:
And the current `low` state isn't sufficient?

mwalliczek:
The current state "low" was wrong implemented (not part of the parsing switch, which could let to errors at startup), missing in the documentation and misleading in its name. It was my fault when I implemented it 1,5 years ago and that's why I want to fix it now.

CurlyMoo:
Was there a unittest for this protocol?

mwalliczek:
No, there was no unit test, because it is difficult to get real data. But I created PR #458 for the rewrite branch anyway.

libs/pilight/protocols/433.92/kerui_d026.c
```diff
@@ -92,6 +95,7 @@ void keruiD026Init(void) {
 	options_add(&kerui_D026->options, "t", "opened", OPTION_NO_VALUE, DEVICES_STATE, JSON_STRING, NULL, NULL);
 	options_add(&kerui_D026->options, "f", "closed", OPTION_NO_VALUE, DEVICES_STATE, JSON_STRING, NULL, NULL);
 	options_add(&kerui_D026->options, "a", "tamper", OPTION_NO_VALUE, DEVICES_STATE, JSON_STRING, NULL, NULL);
+//	options_add(&kerui_D026->options, "b", "battery", OPTION_NO_VALUE, DEVICES_VALUE, JSON_STRING, NULL, NULL);
```
CurlyMoo: The config fails because this line is commented 👍 Remove it and it works fine

CurlyMoo:
Can you add the battery option to the config example in the manual?

libs/pilight/protocols/433.92/kerui_d026.c
mwalliczek: It fails when I try to start it with an existing config without the battery entry ("/devices.c #1443 ERROR: config device setting "battery" of "XY", missing ERROR: failed to read config").
So I think this change would require to manually migrate an existing config and add this entry.

CurlyMoo: True

mwalliczek:
I adapted the battery property to the other protocols and added it to the manual.

CurlyMoo:
Did you have the change to test this in practice?

mwalliczek:
Yes, I tested all changes in practice, including battery: 0, but I was not able to record the raw data for this state for a unit test.

CurlyMoo:
> I was not able to record the raw data for this state for a unit test.

And will you be able to catch it in the near future?

mwalliczek:
> > I was not able to record the raw data for this state for a unit test.
> 
> 
> 
> And will you be able to catch it in the near future?

I'm sorry, but I don't think this is possible, at least not in the next two weeks.

CurlyMoo:
> I'm sorry, but I don't think this is possible, at least not in the next two weeks.

Any estimate when it will be? Because this PR fully focusses on the battery state i would say we at least test it properly.

hugokernel:
> Yes, I tested all changes in practice, including battery: 0, but I was not able to record the raw data for this state for a unit test.

Some door sensor (like wds07) send low battery status without reading the others switch state (door, tamper).

From the implementation of rtl_433 (https://github.com/merbanan/rtl_433/blob/748ac68d54e243b9968d32541f0fa7d973a05985/src/devices/kerui.c), I understand Kerui sensor works like WDS07: Each frame sent from this sensor concern only one state: open / closed / tamper / low.

And therefore that the patch is not correct because we shouldn't have an extra battery field.

We need the raw data to know what's going on (I have a Kerui, I could do some dumps.).

mwalliczek:
I just added a unittest with battery 0 to the rewrite branch, PR #458

PR #459, opened 2020-08-16
Nexus protocol support

ligius-:
I don't see an option to squash the commits, in the PR, hopefully this can be done from your side.

https://docs.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests


ligius-:
I need to redo the PR, seems there is a compilation problem that didn't come up during testing.

libs/pilight/protocols/433.92/nexus.c
```diff
@@ -0,0 +1,253 @@
...
+            temperature += settings->temperature_offset;
+            humidity += settings->humidity_offset;
+            temperature_decimals = (int)round(settings->temperature_decimals);
+            break;
+        }
+        tmp = tmp->next;
+    }
+
+    // format the results into a JSON string
+	snprintf(*message, 255,
```
CurlyMoo: The indenting is not consistent here.

CurlyMoo:
Can you also add some invalid streams?

libs/pilight/protocols/433.92/nexus.c
```diff
@@ -57,7 +57,7 @@ typedef struct settings_t {
     int battery;
     double temperature_offset;
     double humidity_offset;
-    int temperature_decimals;
+    double temperature_decimals;
```
CurlyMoo, line 60: You changed this to double...

libs/pilight/protocols/433.92/nexus.c
```diff
@@ -124,7 +124,7 @@ static void parseCode(char **message) {
             // store or apply the settings
             temperature += settings->temperature_offset;
             humidity += settings->humidity_offset;
-            temperature_decimals = settings->temperature_decimals;
+            temperature_decimals = (int)round(settings->temperature_decimals);
```
CurlyMoo, line 127: But casting to `int` here. Why?

ligius-:
> Can you also add some invalid streams?

The first two tests are invalid streams.
I looked through my capture log, didn't find anything that was corrupted mid-sequence, everything was either rejected from the first or second check, or made it all the way through.
I can create some artificial sequences, I have no problem doing that.

CurlyMoo:
> The first two tests are invalid streams.

Only the first one has a `-1`.

>  I can create some artificial sequences, I have no problem doing that.

Yes please.

libs/pilight/protocols/433.92/nexus.c
ligius-: In the initial PR (not rewrite) this was a double. I changed it to int, but had a compiler warning because json_find_number seems to return a double, so I reverted it. However, I need an int, since that's what makes sense and that's what the printf %.*f needs as an argument for the number of decimals.
I've looked at other protocols and did not see what they do with this number. So I left it as a double but made it actually do something.

ligius-:
> > The first two tests are invalid streams.
> 
> Only the first one has a `-1`.

I even wrote some comments. The first sequence results in the early rejection of the stream, it's the basic sanity check from validate(). The second sequence results in a rejection inside the parse() step. There are 4 additional checks inside parseCode, the last 2 are a kind of crude CRC.

libs/pilight/protocols/433.92/nexus.c
CurlyMoo: True

CurlyMoo:
> The second sequence results in a rejection inside the parse() step.

Which doesn't lead to an error code.

ligius-:
> > The second sequence results in a rejection inside the parse() step.
> 
> Which doesn't lead to an error code.

How do I return an error code from "void parseCode()" ? With logprintf(LOG_ERR,...) ? 
I can move the whole validation logic into "int validate(void)" but that doesn't make much sense to me, since I need to first decode (parse) the signal before I can verify the CRC.

In the test I'm just validating that an empty message is returned, which basically means "no harm done".

CurlyMoo:
> How do I return an error code from "void parseCode()" ?

I was supporting your initial thoughts 😄

PR #440, opened 2019-07-20
Adding IWDS07 protocol (door/window sensor)

1technophile:
Linked to #438 
And following this discussion
https://forum.pilight.org/showthread.php?tid=2952

PR #460, opened 2020-09-02
Fix gs-iwds07 protocol behavior

hugokernel:
Hi,

Thank you for your work on this project.

Next to https://github.com/pilight/pilight/pull/440 from @1technophile, I do some test to figure out how the sensor works.

- When the frame tamped or low is sent from the device, the state of the door is not read.
- The low battery frame (when voltage drop to 1V) is sent when the threshold is reached: the battery status is not read each time the device sent a frame (I think it would be too expensive in energy).

An interesting thread on this sensor https://community.home-assistant.io/t/sonoff-dw1-433mhz-door-window-sensor-low-battery-warning-capable/43073/3


CurlyMoo:
Can you update or create a/the accompianied rewrite PR as well?

hugokernel:
> Can you update or create a/the accompianied rewrite PR as well?

You want only 1 commit, right ?

CurlyMoo:
> You want only 1 commit, right ?

I meant the rewrite branch.

PR #463, opened 2021-02-07
fix(webserver): enable tcp backlog

akloeckner:
This enables the tcp backlog in the webserver's `listen()` call.
The setting of `3` is copied from the core `socket.c`.
https://github.com/pilight/pilight/blob/8b10bfc790bc49960f7f3d3145aadf3446dcba4b/libs/pilight/core/socket.c#L145
See also https://forum.pilight.org/showthread.php?tid=3698 .

### DIFF
--- a/inc/defines.h.in
+++ b/inc/defines.h.in
@@ -52,6 +52,8 @@
 	#define MQTT_ENABLE						1
 #endif
 
+#define TCP_BACKLOG							511 // comparable to apache, nginx, etc.
+
 #define MAX_CLIENTS							30
 #define BUFFER_SIZE							1025
 #define WIRINGX_BUFFER					4096
--- a/libs/pilight/core/mqtt-server.c
+++ b/libs/pilight/core/mqtt-server.c
@@ -883,7 +883,7 @@ int mqtt_server(int port) {
 		/*LCOV_EXCL_STOP*/
 	}
 
-	if((listen(sockfd, 3)) < 0) {
+	if((listen(sockfd, TCP_BACKLOG)) < 0) {
 		/*LCOV_EXCL_START*/
 		logprintf(LOG_ERR, "listen: %s", strerror(errno));
 #ifdef _WIN32
--- a/libs/pilight/core/mqtt-server.c
+++ b/libs/pilight/core/mqtt-server.c
@@ -883,7 +883,7 @@ int mqtt_server(int port) {
 		/*LCOV_EXCL_STOP*/
 	}
 
-	if((listen(sockfd, 0)) < 0) {
+	if((listen(sockfd, 3)) < 0) {
 		/*LCOV_EXCL_START*/
 		logprintf(LOG_ERR, "listen: %s", strerror(errno));
 #ifdef _WIN32
--- a/libs/pilight/core/socket.c
+++ b/libs/pilight/core/socket.c
@@ -141,8 +141,7 @@ int socket_start(unsigned short port) {
 	}
 
 	int x = 0;
-	//try to specify maximum of 3 pending connections for the master socket
-	if((x = listen(socket_server, 3)) < 0) {
+	if((x = listen(socket_server, TCP_BACKLOG)) < 0) {
 		logprintf(LOG_ERR, "failed to listen to socket");
 		exit(EXIT_FAILURE);
 	}
--- a/libs/pilight/core/webserver.c
+++ b/libs/pilight/core/webserver.c
@@ -1989,7 +1989,7 @@ static int webserver_init(int port, int is_ssl) {
 		return -1;
 	}
 
-	if((listen(sockfd, 3)) < 0) {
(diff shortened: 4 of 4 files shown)


CurlyMoo:
Does it fix the synfloading as well?

akloeckner:
Yes. I am able to connect and there is no SYN flood warning in `dmesg` anymore.

In case we want to be super proper: I am seeing new warnings for port `1883` now:
```
[403657.700489] TCP: request_sock_TCP: Possible SYN flooding on port 1883. Dropping request.  Check SNMP counters.
```
But this is not related to me trying to open the webpage or connecting through `pilight-receive`. Google suggests it is some MQTT service, which I have no idea why it should be running in the container. Maybe this port number rings a bell with you.

Also, I cannot connect to `https` on port `5002` currently, But I presume this is related to something different. I think, I had never tried to do that before, anyways.

CurlyMoo:
Yes, 1883 is the internal MQTT server. Can you check if that listen patch fixes this error as well?

akloeckner:
Yes, sure. Where do I need to change that? (I can't find anything, if I search for mqtt in your repo...)

CurlyMoo:
https://github.com/pilight/pilight/blob/staging/libs/pilight/core/mqtt-server.c

akloeckner:
Hmm... I must have been blind. I tried, and it does fix that, too:

Before I had the synflood warning when restarting the service. With the fix, that warning vanishes.

CurlyMoo:
Let it run for a few days to be sure,

akloeckner:
Alright!

akloeckner:
The patch is running without any problems so far in my productive setup (6 days in a row without restart after finally setting everything up and until I came back to it). I can connect to the webservers and see data there. There has not been any warning for the MQTT service anymore.

I still seem to see the warning for the first connection to the web servers after a restart. However, the page loads fine regardless.

Not sure about the latter. I could try and dig deeper. But, actually, a working server with a warning appears better to me than a not-working one. If I were to dig deeper, I'd probably even increase the backlog to `511`, which is apparently what servers like nginx or apache use by default.

CurlyMoo:
> But, actually, a working server with a warning appears better to me than a not-working one.

True

> I'd probably even increase the backlog to 511, which is apparently what servers like nginx or apache use by default.

Worth a try!

akloeckner:
Ok. I'll dig deeper as soon as I find the time .Since it seems to turn out as a tuning parameter: where would I most meaningfully `def`ine this backlog as a constant?

CurlyMoo:
>where would I most meaningfully define this backlog as a constant?

in `inc/defines.h.in`

akloeckner:
> I still seem to see the warning for the first connection to the web servers after a restart.

Seems to be fixed. I'll have it running for the week and report back...

akloeckner:
Everything looks fine. No warnings, interface accessible. Let me know, if there's anything you need me to change still!

CurlyMoo:
A simple file content search gave me this additional list where `listen` is used:
- coap.c
- mdns.c
- socket.c

akloeckner:
> * coap.c
https://github.com/pilight/pilight/blob/8b10bfc790bc49960f7f3d3145aadf3446dcba4b/libs/pilight/core/coap.c#L339

This is not a plain `listen()` call.
 
> * mdns.c
https://github.com/pilight/pilight/blob/8b10bfc790bc49960f7f3d3145aadf3446dcba4b/libs/pilight/lua_c/network/mdns.c#L1039
https://github.com/pilight/pilight/blob/8b10bfc790bc49960f7f3d3145aadf3446dcba4b/libs/pilight/lua_c/network/mdns.c#L1051

Neither are those.

> * socket.c

https://github.com/pilight/pilight/blob/8b10bfc790bc49960f7f3d3145aadf3446dcba4b/libs/pilight/core/socket.c#L145

This one already has a TCP backlog setting of `3`. Do you want me to change this one to `511` then, too? I don't expect that to have negative side effects. In the opposite: pilight will be able to handle more than 3 connection attempts at the time then. However, I am be no means an expert at this level of networking. And I am unsure how I would test this. Other than letting it run for a few days again...

Edit: note that this line has been around for ages. I couldn't find a commit detailing the choice of three as a backlog. So, I guess it won't harm to increase. If an operating system has less resources, it should overwrite this value anyways. 

https://github.com/pilight/pilight/blob/17532a3dd78ec27b3b56f7b71c78b64cf563c396/libs/socket.c#L68

CurlyMoo:
> Do you want me to change this one to 511 then, too?

Yes

CurlyMoo:
Can you do the same commits in rewrite?

akloeckner:
Not sure what you mean. Port the change to the rewrite branch? I'll see. Should be straight forward, I guess?

CurlyMoo:
Yes. I think it's fair to keep this fix attributed to your name.

akloeckner:
That's nice of you to think about. :-) I created the PR #467.

PR #469, opened 2022-11-25
Fix build for ESP8266

davidfurey:
## Declare global variables extern in header files
And define them in corresponding c files. GCC 10.1.0 changes the default compilation options to use -fno-common,
which means that by default, this project no longer links unless you override the default with -fcommon.

## Remove unnecessary include files
`backtrace.h` is included in `mem.c` only when DEBUG is enabled, I cannot see why they should always be included in `mem.h`.  In the context of ESPilight, `backtrace.h` isn't available.

~These changes allow the ESPilight build to complete in the context of OpenMQTTGateway, but I'm marking as a draft until I've tested it on an ESP8266~ Works for me on ESP8266

### DIFF
--- a/libs/pilight/config/gui.h
+++ b/libs/pilight/config/gui.h
@@ -46,7 +46,7 @@ struct gui_elements_t {
 	struct gui_elements_t *next;
 };
 
-struct config_t *config_gui;
+extern struct config_t *config_gui;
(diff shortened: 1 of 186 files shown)


davidfurey:
Closing due to inactivity

